PR #22, opened 2026-08-19
Practice stats and journal

### LINKED ISSUE #1: Practice stats and journal
Track practice activity per piece so the library reflects what is actually being worked on.

- Session logging: time spent per score, last-practiced timestamp, session count
- Surface `last practiced` and total time on the score detail page
- `Recently practiced` and `Neglected` library filters
- Extends the existing `last_page` per-score state in the `scores` table

sirged:
## Summary
- Adds a `practice_sessions` table (score_id, started_at, seconds, note) to track practice activity per score.
- New endpoints: log a practice session, fetch a score's session history + totals, and a library-wide weekly summary with top-practiced scores.
- `GET /api/scores` now returns `last_practiced` and `practice_seconds` per score (single aggregate query, no N+1) and supports `practiced=recent` / `practiced=neglected` filters.
- Library sidebar gains "Recently practiced" and "Needs attention" filters, and score cards show a subtle "practiced Nd ago" line.

No timer UI in the viewer — this is the data layer plus library surfacing; the viewer timer is a later issue.

Closes #1

## Test plan
- [x] `npm run build` in `web/` — succeeds with zero errors
- [x] Python import sanity check against `fermata.api` / `fermata.db` in a clean container — succeeds

### DIFF
--- a/server/fermata/api.py
+++ b/server/fermata/api.py
@@ -4,7 +4,7 @@ from pathlib import Path
 
 from fastapi import APIRouter, HTTPException, UploadFile
 from fastapi.responses import FileResponse
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from . import scanner
 from .config import FILE_TYPES, LIBRARY_DIR
@@ -14,6 +14,7 @@ from .thumbs import thumb_path
 router = APIRouter(prefix="/api")
 
 VALID_KINDS = {"notation", "tab", "both", "unknown"}
+VALID_PRACTICED = {"recent", "neglected"}
 
 
 def _score_row(conn, score_id: int):
@@ -69,6 +70,8 @@ def list_scores(
     favorite: bool = False,
     practiced: str = "",
 ):
+    if practiced and practiced not in VALID_PRACTICED:
+        raise HTTPException(422, f"practiced must be one of {sorted(VALID_PRACTICED)}")
     conn = connect()
     sql = "SELECT DISTINCT s.* FROM scores s"
     where, params = [], []
@@ -192,12 +195,12 @@ def patch_score(score_id: int, patch: ScorePatch):
 
 class PracticeIn(BaseModel):
     seconds: int
-    note: str | None = None
+    note: str | None = Field(default=None, max_length=2000)
 
 
 def _practice_totals(conn, score_id: int):
     row = conn.execute(
-        """SELECT COUNT(*) AS session_count, COALESCE(SUM(seconds), 0) AS total_seconds,
+        """SELECT COUNT(*) AS session_count, COALESCE(SUM(seconds), 0) AS practice_seconds,
                   MAX(started_at) AS last_practiced
            FROM practice_sessions WHERE score_id = ?""",
         (score_id,),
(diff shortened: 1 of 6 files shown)


sirged:
Pushed fixes for all findings, merged on top of latest main (which already carried per-file scan error handling and practice-setting reapplication — kept both).

**1. Rename-safety data loss (critical)** — `scanner.py`: when a path has no matching row, the scanner now looks for an existing row with the same content `hash` whose old path is no longer on disk (computed once up front as `disk_paths`, so this is independent of scan order). Exactly one match → `UPDATE` that row's path/metadata in place, preserving `id` and everything foreign-keyed to it (`practice_sessions` via `ON DELETE CASCADE`). Zero or multiple matches → falls back to the original insert/stale-row-delete behavior, since with genuine duplicate content there's no way to know which physical file moved. Verified with a script that scans a library, logs a practice session, renames the file, rescans, and asserts the session is still attached to the same score id at the new path — passed. Also verified the multi-duplicate case (two identical files, rename only one): only the renamed one gets re-linked; the untouched duplicate and both sessions stay correctly separated, since its old path is still present in `disk_paths` and it's excluded as a candidate.

**2. No writer for practice_sessions** — added a start/stop practice timer to the score header in `Viewer.svelte` (mm:ss while running, `.ghost` button styling matching the rest of the header). On stop it POSTs elapsed seconds via `logPractice`; sessions under 10s are discarded. A running session is flushed (or discarded if too short) on switching to a different score, on unmount, and on `beforeunload`, so it's never silently lost.

**3. Weekly/all-time mismatch** — `top_scores` in `/practice/summary` now scopes to `started_at >= datetime('now','-7 days')`, matching `week_seconds`/`week_sessions`.

**4. Field naming** — `/scores/{id}/practice`'s aggregate is now `practice_seconds` (was `total_seconds`), matching the score list/detail payloads.

**5. Unstable ordering** — session list query is now `ORDER BY started_at DESC, id DESC`.

**6. Unvalidated `practiced` param** — added a `VALID_PRACTICED` allow-list (same pattern as `VALID_KINDS`); unknown non-empty values now 422, empty string still means "no filter".

**7. Unbounded note** — `PracticeIn.note` capped at 2000 chars.

**8. "practiced NaNd ago" / off-by-a-day** — `Library.svelte` now compares calendar dates instead of flooring raw elapsed milliseconds, phrases same-day as "practiced <24h ago", and bails out to an empty string on an unparseable timestamp.

`npm run build` is clean (zero errors). Verification test output:

```
== Phase 1: rename-safety ==
scanned original: id=1 path=Test Piece.musicxml
practice session logged
rows after rename+rescan: [(1, 'Renamed Piece.musicxml')]
PASS: practice session survived rename, still attached to the same score id at the new path

== Phase 2: duplicate-safety ==
scanned duplicates: Dup One id=2, Dup Two id=3
duplicate rows after rename+rescan: [(2, 'Dup One Renamed.musicxml'), (3, 'Dup Two.musicxml')]
PASS: renaming one duplicate re-linked only that row; the untouched duplicate and both practice histories stayed correct

ALL TESTS PASSED
```

PR #23, opened 2026-08-19
Gig mode: distraction-free performance view

### LINKED ISSUE #5: Gig mode: distraction-free performance view
A performance mode for a tablet on a music stand.

- Fullscreen with no UI chrome
- Screen wake lock so the display never sleeps mid-piece
- Oversized pedal/tap targets and half-page turns
- Works for both the PDF reader and the interactive viewer

sirged:
## Summary

Adds a distraction-free "gig mode" for using the viewer on a tablet on a music stand mid-performance.

- Toggle from the viewer header, the `f` key, or `Escape` to exit. The header/toolbar hides entirely while active.
- Requests browser fullscreen on the viewer container and exits it when leaving gig mode; also listens for `fullscreenchange` so exiting fullscreen via the browser's own UI (or Escape) keeps gig-mode state in sync instead of getting stuck.
- Requests a screen wake lock (`navigator.wakeLock`) while active so the display doesn't sleep mid-piece. Feature-detected, degrades silently where unsupported, released on exit/teardown, and re-acquired on `visibilitychange` since the OS drops wake locks when a tab is backgrounded.
- PdfViewer gets oversized, mostly-invisible left/right tap zones for page turns (foot pedal / thumb friendly), with a subtle highlight on hover/press.
- Adds a half-page advance mode (defaults on when entering gig mode, toggleable) so the eye doesn't lose its place on a full page jump; existing arrow-key/space page-turn behavior keeps working both in and out of gig mode.
- Dark-mode inversion setting is untouched by gig mode - the HUD hides but the toggle and its state persist.

Closes #5

## Test plan

- [x] npm run build in web/ succeeds with zero errors
- [ ] Manual: toggle gig mode with the button, f, and Escape; confirm fullscreen enter/exit and header hide/show
- [ ] Manual: exit fullscreen via the browser (not the app) and confirm gig mode state resyncs
- [ ] Manual: verify screen stays awake during gig mode on a device that supports Wake Lock, and that unsupported browsers don't throw
- [ ] Manual: tap zones and half-page toggle on a touch device

### DIFF
--- a/server/fermata/scanner.py
+++ b/server/fermata/scanner.py
@@ -14,6 +14,8 @@ _state = {
     "added": 0,
     "updated": 0,
     "removed": 0,
+    "errors": 0,
+    "last_error": None,
     "started_at": None,
     "finished_at": None,
 }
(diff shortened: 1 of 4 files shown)


sirged:
Pushed fixes for all ten review findings (the original eight plus the two from the follow-up adversarial pass):

1. **Ctrl+F/Cmd+F hijack** — `onKey` now bails out early when `ctrlKey`/`metaKey`/`altKey` is held, so browser find is untouched.
2. **No exit path for non-PDF gig mode** — `gigMode`/`onToggleGig` are now passed into `TabViewer`. In gig mode it swaps its full practice toolbar for a minimal floating HUD (play/pause, stop, exit) — always reachable by touch alone, no keyboard needed.
3. **Orphaned wake lock on fast exit** — added a `wantWakeLock` intent flag; if gig mode ends while `wakeLock.request()` is still in flight, the resolved lock releases itself instead of pinning the screen awake.
4. **Wake lock re-acquire/release races** — `acquireWakeLock` now returns early if a lock is already held, and the `release` listener compares sentinel identity before nulling `wakeLock`, so a stale release can't clobber a newer lock.
5. **PDF pages stale at old width in fullscreen** — added a debounced `ResizeObserver` on the scroller that re-renders existing canvases in place at the new width, then restores scroll to the current page. IntersectionObserver-driven page tracking is suppressed for the duration of a resize pass so it can't record a bogus page mid-reflow.
6. **Tap zones blocking scroll/scrollbar** — replaced the absolutely-positioned overlay buttons with pointerdown/up handlers on the scroller itself; a "tap" is only recognized after the fact from how little the pointer moved, so real drags/scrolls (and the native scrollbar) are never intercepted.
7. **Space/arrows swallowed while typing tags** — `PdfViewer`'s key handler now has the same `INPUT`/`TEXTAREA`/`SELECT`/`isContentEditable` guard `Viewer` already had.
8. **Half-page drift** — the turn distance is now derived from the current `.pdf-page` element's actual rendered height instead of `container.clientHeight`.
9. **No `e.repeat` guard + async interleaving** — added an `e.repeat` early return, and gig mode enter/exit is now reentrancy-safe via a generation counter: a stale `enterGigMode` continuation (e.g. F then Escape fired before `requestFullscreen` resolved) can no longer leave the browser stuck in real fullscreen with `gigMode` reading false, nor acquire a wake lock it shouldn't.
10. **IntersectionObserver picking the wrong page** — the callback now picks the entry with the highest `intersectionRatio` among intersecting entries instead of whichever is last in the batch, so half-page turns land the tracked/saved page correctly.

Touch-only exit is confirmed reachable in both viewers: PdfViewer's HUD always renders its exit button in gig mode (not conditionally hidden), and TabViewer's new gig HUD does the same.

`npm run build` passes clean (zero errors/warnings). Merged latest `main` first; no `server/` files touched.

sirged:
Merged latest main (PR #22's practice timer) and resolved the Viewer.svelte conflict.

**Conflict resolution:** it was a pure markup conflict in the header — the actual timer logic (`practiceStart`/`practiceElapsed`/`startPractice`/`flushPractice`, the score-switch flush effect, and the `beforeunload` flush) had already merged cleanly and is untouched. Only the `<div class="controls">` block needed reconciling; it now renders tags editing → content-kind select → practice timer button → favorite → Gig mode button, preserving both features exactly as each branch had them.

**Timer + gig mode interaction:** kept the timer running through gig mode rather than stopping it on entry — a performance is practice time, and stopping it would silently truncate whatever session was already running. Since `enterGigMode` never touched practice state to begin with, this required no logic change. What it did need: the timer lives in the header, which gig mode hides, so a running session would have gone invisible with no way to see or stop it. Fixed by threading a derived `practiceLabel` (formatted mm:ss, or `null` when not running) and `onStopPractice` down into both `PdfViewer` and `TabViewer`. Their gig HUDs now show a small `● mm:ss` indicator next to the exit button whenever a session is active, tappable to stop/log it without leaving gig mode. Nothing shows when no session is running.

`npm run build` passes clean (zero errors/warnings). `git diff --stat origin/main...HEAD` shows only the three web/src/lib viewer files — no server/ changes.

PR #33, opened 2026-08-19
Side-by-side PDF and staff view with transcription editing

sirged:
## Summary
- Adds a layout switcher to the viewer for PDF scores: PDF only, staff only, or side by side (PDF left, staff right, stacking vertically on narrow screens). Choice persists in localStorage, defaulting to side-by-side the first time a transcription exists.
- Empty state in the staff pane invites transcription via `POST /transcribe`, with an optional (clearly-labeled) time-signature hint since auto-detection is unreliable.
- Extraction warnings (e.g. "durations inferred from spacing", "time signature not detected") are surfaced as a prominent caution banner in the staff pane, never buried.
- Minimal alphaTex source editor: view/edit the transcription, save via `PUT`, staff re-renders from the saved content, with a "Revert to extracted" action (re-runs `POST /transcribe`) when an edit goes wrong.
- Degrades gracefully if the transcription endpoints 404 or the backend isn't available yet — no faked data, just a clear state and a retry action.

## Implementation
- `web/src/lib/api.js`: `transcription`, `transcribe`, `saveTranscription`, `transcriptionAnalysis` bindings.
- `web/src/lib/TabViewer.svelte`: now accepts a `tex` prop (alphaTex content) alongside `score`/`demo`, so it can render a transcription directly and is reused as-is for the staff pane — no second alphaTab integration. Existing gig mode, playback, loop, metronome, count-in and tempo ladder all still apply. Also clears stale parse errors on each new load.
- `web/src/lib/ScoreCompare.svelte` (new): owns layout state/persistence, transcription fetch/transcribe/save/revert, the warnings banner, and composes `PdfViewer` + `TabViewer` into the three layouts. Swapped in for `PdfViewer` in `Viewer.svelte` for PDF scores; non-PDF scores are untouched.

## Test plan
Backend endpoints aren't merged in this branch, so verification stubbed the network via Playwright `page.route` against a `vite preview` build (25/25 assertions passed headlessly, chromium):
- [x] Demo route regression: alphaTab still renders non-zero-size SVG, no console/page errors (confirms the recent alphaTab Vite plugin fix isn't regressed)
- [x] Graceful degradation: `/transcription` 404 → PDF pane still renders, staff pane shows the "No staff transcription yet" empty state with the Transcribe CTA, no fake data, no unexpected console errors
- [x] Realistic transcription payload → side-by-side is the default the first time a transcription exists; both PDF canvas and staff SVG are non-zero size; warnings banner shows both warning strings verbatim, styled as caution
- [x] Layout switching (PDF / Staff / Side by side) hides/shows the right panes; last explicit choice persists across reload
- [x] Editor round-trip: opens pre-filled, saves through `PUT`, source badge flips to "edited"; "Revert to extracted" re-runs `POST /transcribe` and flips it back

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017T6Jz8bFWsowh55GKfBv6i

### DIFF
--- a/web/src/lib/ScoreCompare.svelte
+++ b/web/src/lib/ScoreCompare.svelte
@@ -1,0 +1,487 @@
+<script>
+  import { untrack } from "svelte";
+  import { api } from "./api.js";
+  import PdfViewer from "./PdfViewer.svelte";
+  import TabViewer from "./TabViewer.svelte";
+
+  let {
+    score,
+    gigMode = false,
+    onToggleGig = () => {},
+    practiceLabel = null,
+    onStopPractice = () => {},
+  } = $props();
+
+  const LAYOUT_KEY = "fermata.layout";
+
+  function loadStoredLayout() {
+    try {
+      return localStorage.getItem(LAYOUT_KEY);
+    } catch {
+      return null;
+    }
+  }
+
+  function persistLayout(l) {
+    try {
+      localStorage.setItem(LAYOUT_KEY, l);
+    } catch {
+      // storage unavailable (private browsing, etc) - layout just won't persist
+    }
+  }
+
+  // no stored preference yet: lean on side-by-side as soon as there's
+  // something to compare, since verification is the whole point
+  let layout = $state(loadStoredLayout() ?? (untrack(() => score.has_transcription) ? "side" : "pdf"));
+
+  function setLayout(l) {
+    layout = l;
+    persistLayout(l);
+  }
+
+  function maybeDefaultToSide() {
+    if (loadStoredLayout()) return; // user already made an explicit choice
+    layout = "side";
+  }
+
+  // "loading" | "none" | "transcribing" | "ready" | "error"
+  let transcriptionState = $state("loading");
+  let transcription = $state(null);
+  let fetchError = $state("");
+
+  let tsNum = $state("");
+  let tsDen = $state("");
+
+  let editorOpen = $state(false);
+  let draft = $state("");
+  let saving = $state(false);
+  let saveError = $state("");
+  let reverting = $state(false);
+
+  async function loadTranscription() {
+    transcriptionState = "loading";
+    fetchError = "";
+    try {
+      const t = await api.transcription(score.id);
+      transcription = t;
+      draft = t.content;
+      transcriptionState = "ready";
+      maybeDefaultToSide();
+    } catch (e) {
+      const msg = String(e?.message ?? e);
+      if (msg.startsWith("404")) {
+        transcriptionState = "none";
+      } else {
+        // endpoint missing entirely, network down, backend not merged yet -
+        // degrade rather than pretend a transcription exists
+        transcriptionState = "error";
+        fetchError = msg;
+      }
+    }
+  }
+
+  $effect(() => {
+    void score.id;
+    loadTranscription();
+  });
+
+  async function runTranscribe() {
+    transcriptionState = "transcribing";
+    fetchError = "";
+    const n = Number(tsNum);
+    const d = Number(tsDen);
+    const body = {};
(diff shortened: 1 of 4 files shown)


sirged:
Pushed fixes for the review findings on the frontend side (`web/**`) of this PR. Backend findings (DELETE endpoint, etc.) are being handled separately.

**1. Revert didn't actually revert (HIGH)** — `revertToExtracted()` now calls the new `DELETE /api/scores/{id}/transcription` (`api.deleteTranscription`) instead of re-running `/transcribe`. A 404 (nothing left after the edit was removed) shows a clear "Reverted — no extracted transcription was left to fall back to" message instead of pretending the edit is still there; a 405 (DELETE not deployed yet) says so explicitly rather than silently leaving the stale edit in place.

**2. Unsaved editor text discarded on unrelated metadata edits (HIGH)** — `ScoreCompare.svelte` now tracks `let scoreId = $derived(score.id)` and keys the load effect off that instead of the whole `score` object. Also had to route `loadTranscription`'s own `api.transcription(...)` call through that same `scoreId` value — reading `score.id` directly inside the function was enough to re-attach the effect to the whole object (it runs synchronously up to the first `await`, inside the effect's tracking window) and defeated the fix on the first pass. Verified: editing the draft then toggling ★ or the content-type dropdown no longer reloads/wipes it.

**3. Warnings banner disappeared on reload (HIGH)** — now reads `transcription.warnings ?? confidence.warnings`, tolerant of `confidence` still being a raw JSON string. Confirmed against `server/fermata/api.py` on `feature/tab-extraction`: `POST /transcribe` echoes `warnings` at the top level, `GET /transcription` only nests it under `confidence.warnings`.

**4. No request cancellation (MEDIUM)** — added a generation counter; `loadTranscription`/`runTranscribe` bail if superseded by a newer call, and the effect's cleanup bumps it too.

**5. Player transport state stale after rebuild (MEDIUM)** — `TabViewer.svelte` now resets `playerReady`/`playing` alongside `loadError` when the player is rebuilt.

**6. Layout switch destroyed/remounted both panes (MEDIUM)** — both panes now always render; the inactive one is pulled out of flow with `position: absolute; inset: 0; visibility: hidden` (not `display: none`, so a pane mounted while hidden still measures a real width). PdfViewer and TabViewer instances persist across layout switches.

**7. Gig mode showed two HUDs (MEDIUM)** — added `activeLayout`, which collapses `side` to a single pane while `gigMode` is on, so exactly one viewer (and its HUD/exit control) is visible.

**8. API errors lost the server's `detail` (MEDIUM)** — `api.js`'s `j()` now parses and surfaces FastAPI's `detail`, with `res.statusText` only as a last-resort fallback (empty over HTTP/2). Added `ApiError` carrying `.status`; the 404 branches in `ScoreCompare.svelte` now check `e.status` instead of string-sniffing the message.

**9. Layout default guard broken without localStorage (LOW)** — added an in-memory `userChoseLayout` flag alongside the persisted one.

**10. Revert re-extracting without time signature (LOW)** — resolved by #1: revert no longer calls `/transcribe` at all.

**11. Save allowed with empty content (LOW)** — `saveEdit()` short-circuits with a message, and the Save button is disabled, when `draft.trim()` is empty.

**12. Unused analysis call (LOW)** — wired `transcriptionAnalysis` into the empty state: if it reports `extractable: false`, the copy explains why instead of offering a "Transcribe this PDF" button that can only fail.

**Verification** — extended the Playwright check (stubbing `/api` via `page.route`, run against `vite preview`, no backend needed): 14/14 assertions pass, covering the warnings-banner shape, draft surviving unrelated metadata edits, layout switching not re-fetching the PDF or the soundfont, gig mode showing exactly one HUD/exit control, revert calling DELETE and surfacing a message on a stubbed 404, and the `#/demo` route still rendering with zero console errors.

PR #34, opened 2026-08-19
Extract guitar tablature from engraved PDFs

sirged:
## Summary

Adds server-side tablature extraction: a PDF whose pages carry a real 6-line
tab staff (Finale/Sibelius/Opus-style vector engraving) can be turned into
alphaTex and stored against the score.

- `server/fermata/tabextract.py` - staff systems are located from vector
  line primitives (6 evenly-spaced long horizontal lines = tab staff, 5 =
  standard staff); fret numbers come from digit-only text spans assigned to
  the nearest tab staff/string by position, not by font (different
  engravers use different fonts for fret numbers). Barlines come from
  vertical line primitives. `analyze()` is a cheap triage (vector vs.
  raster, staff counts); `extract()` does the full pass and returns an
  `ExtractionResult` (alphaTex, bars/beats/notes, tuning, per-section
  confidence, warnings). Neither function ever raises - a malformed PDF or
  one with no tab staves comes back as `extractable: false` with a reason.

- `transcriptions` table (`server/fermata/db.py`) - one `source='extracted'`
  row and at most one `source='edited'` row per score, enforced by a unique
  index on `(score_id, source)`. Re-running extraction only ever
  `INSERT ... ON CONFLICT` upserts the `extracted` row, so a hand-edited
  transcription can never be silently overwritten by a re-extraction.
  `GET .../transcription` prefers the edited row when both exist.

- New endpoints on the existing `/api` router:
  - `GET /api/scores/{id}/transcription/analysis` - cheap triage result
  - `POST /api/scores/{id}/transcribe` - runs extraction, accepts an
    optional `{time_signature: [num, den]}` override, stores + returns the
    extracted row plus warnings
  - `GET /api/scores/{id}/transcription` - current transcription (edited
    wins) or 404
  - `PUT /api/scores/{id}/transcription` - save hand-edited alphaTex
    (`source='edited'`)
  - `GET/api/scores` and `GET /api/scores/{id}` now include a batched
    `has_transcription` boolean (extends the existing `_with_tags` helper,
    no N+1 query)

- **Sync, not background-thread**: measured full extraction (both pages of
  a real 2-page tab score) at ~0.06-0.2s including DB round trip, so
  `POST /api/scores/{id}/transcribe` runs synchronously rather than reusing
  `scanner.py`'s background-thread/status-polling pattern - there's no
  latency problem to solve.

## Accuracy (validated against real library files)

- Finale-engraved tab PDF (a library score, 2 pages): 5 tab + 5 notation
  staves per page detected, 357 notes across 49 bars extracted, tuning
  correctly read as Drop D, alphaTex parses cleanly.
- Standard-notation-only PDF (Tarrega study, fingering digits present but
  no tab staff): correctly reports `extractable: false` - fingering numbers
  are not mistaken for fret numbers.
- Raster-scanned PDF (Clair de Lune): correctly reports `extractable: false`
  with reason "raster scan" (no fonts/drawings/text on any page).

## Known limitation - please read

**Durations and time signatures are low confidence.** Rhythm is inferred
from horizontal spacing between note columns snapped to the nearest plain
duration - there is no dotted-note or tie model, so it will get simple
passages roughly right and can be wrong wherever engraving spacing doesn't
scale linearly with duration. Time-signature auto-detection is essentially
non-functional: the glyphs live inside subsetted music fonts (Maestro,
Opus, ...) at remapped codepoints that don't correspond to their visual
meaning, so it fails silently on real files. `extract()` accepts a manual
`time_signature` override for exactly this reason, and every extraction
result carries this out explicitly in `warnings` / `confidence` rather than
papering over it - callers (and the UI) should surface these, not hide them.

## Test plan

- [x] `server/tests/` - pytest suite covering: Finale tab PDF extracts
  expected notes/bars (>=185 notes / >=24 bars, floor from the validated
  single-page numbers since this file is 2 pages), notation-only PDF yields
  `extractable: false` with no tab staves, raster PDF reports
  not-extractable, malformed/missing PDF never raises, and an edited
  transcription survives a re-extraction. Uses the real sample PDFs
  read-only via `FERMATA_TEST_LIBRARY`, skips if unset/missing.
- [x] Ran in a clean `python:3.12-slim` container against the real library
  (12 passed).
- [x] Built the full Docker image from this branch and exercised the new
  endpoints against the real library with curl: transcribed a library score (3/4 override), confirmed warnings are present and honest, confirmed
  `has_transcription` flips true, confirmed a `PUT` edit survives a
  follow-up `POST /transcribe`.

https://claude.ai/code/session_017T6Jz8bFWsowh55GKfBv6i

### DIFF
--- a/server/fermata/api.py
+++ b/server/fermata/api.py
@@ -1,20 +1,23 @@
+import json
 import re
 import shutil
 from pathlib import Path
 
-from fastapi import APIRouter, HTTPException, UploadFile
+from fastapi import APIRouter, Body, HTTPException, UploadFile
 from fastapi.responses import FileResponse
 from pydantic import BaseModel, Field
 
 from . import scanner
 from .config import FILE_TYPES, LIBRARY_DIR
 from .db import connect, tx
+from .tabextract import analyze as analyze_pdf, extract as extract_pdf
 from .thumbs import thumb_path
 
 router = APIRouter(prefix="/api")
 
 VALID_KINDS = {"notation", "tab", "both", "unknown"}
 VALID_PRACTICED = {"recent", "neglected"}
+MAX_TRANSCRIPTION_CHARS = 2_000_000
 
 
 def _score_row(conn, score_id: int):
@@ -28,6 +31,7 @@ def _with_tags(conn, rows):
     ids = [r["id"] for r in rows]
     tag_map: dict[int, list[str]] = {i: [] for i in ids}
     practice_map: dict[int, dict] = {}
+    transcribed_ids: set[int] = set()
     if ids:
         placeholders = ",".join("?" * len(ids))
         for r in conn.execute(
@@ -46,11 +50,17 @@ def _with_tags(conn, rows):
                 "practice_seconds": r["practice_seconds"],
                 "last_practiced": r["last_practiced"],
             }
+        for r in conn.execute(
+            f"""SELECT DISTINCT score_id FROM transcriptions WHERE score_id IN ({placeholders})""",
+            ids,
+        ):
+            transcribed_ids.add(r["score_id"])
     out = []
     for r in rows:
         d = dict(r)
         d["favorite"] = bool(d["favorite"])
         d["tags"] = tag_map.get(r["id"], [])
+        d["has_transcription"] = r["id"] in transcribed_ids
         d.update(practice_map.get(r["id"], {"practice_seconds": 0, "last_practiced": None}))
         out.append(d)
     return out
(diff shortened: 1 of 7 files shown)


sirged:
Pushed fixes for all 10 findings plus the cross-PR DELETE endpoint. Summary:

1. **`\tuning` order (critical)** - alphaTex binds the first `\tuning` entry to string 1, and `string_for_y` assigns string 1 to the top tab line (high string). `_build_alphatex` now emits `reversed(tuning)`, so a high-string-first line is written while `result.tuning`/`DEFAULT_TUNING`/`DROP_D_TUNING` stay low-to-high everywhere else. Verified against alphaTab's own `AlphaTexImporter`: with the fixed emission, `\tuning E4 B3 G3 D3 A2 E2` + `0.1` parses to MIDI 64 (high E) and `0.6` parses to MIDI 40 (low E); the old low-to-high order gave `0.1` a MIDI pitch of 40 (the mirrored string). Also added `test_tuning_emitted_high_string_first_for_alphatex_binding`.

2. **6/8 duration budget** - added `_measure_quarter_length(ts)` (`ts[0] * 4 / ts[1]`) and used it everywhere instead of `float(ts[0])`, so 6/8 now budgets 3.0 quarters per measure, same as 3/4. New tests: a direct table-driven test on `_measure_quarter_length`, plus `test_six_eight_and_three_four_extract_identically` against the real Zanarkand fixture (asserts identical bar/duration output for `(3,4)` vs `(6,8)`).

3. **Unescaped `\title`** - added `_escape_tex_string` (backslash then quote) and applied it to the title. Subtitle/tuning-name fields were checked too; nothing else interpolates untrusted strings into alphaTex. New tests on `_escape_tex_string` and `_build_alphatex` with a hostile title (`she said "hi"\bye`).

4. **`time_signature` validation** - `api.py` now rejects a numerator outside 1-32 or a denominator that isn't a power of two (1/2/4/8/16/32) with a 422, before it ever reaches the extractor. Parametrized test covers `(0,4)`, `(33,4)`, `(4,0)`, `(4,3)`, `(4,6)`, plus a positive test for `(6,8)`.

5. **Rest-only bars skipped** - empty bars now emit an explicit rest beat (`:<dur> r`) instead of being `continue`d, so bar numbering stays aligned with the PDF. Verified with a synthetic 3-bar PDF (middle bar empty) and against the real Zanarkand file, which does contain a genuine rest bar.

6. **`tab_staff_count` inconsistency** - both `analyze()` and `extract()` now report the same thing: the total tab/standard staff count *summed across pages*, not a per-page max. Documented on `ExtractionResult`. Existing exact-count assertions were loosened to floors (consistent with how `bars`/`notes` are already asserted), and a new synthetic 2-page test pins the sum behavior precisely.

7. **Digit-merge window / impossible frets** - the merge-window average now only considers single-character token widths (excluding already-merged 2-char tokens that were inflating it), and the window itself is tighter (0.35x avg width vs 0.5x). Merges that would still produce a fret above 24 are rejected and kept as separate notes, surfaced via a warning. While testing against the real Zanarkand file I found a *second* path to the same problem: PDFs can contain a native two-character span (e.g. `"26"`) that never goes through the merge logic at all. Extended the same ceiling check to that case too - it's kept (can't safely split an already-single span) but counted and surfaced as a distinct "suspicious fret" warning. Four new tests cover: wide-gap non-merge, valid 2-digit merge, rejected impossible merge, and the native-span case.

8. **`extractable=true` with zero notes** - if tab staves were found but no digit tokens ever matched any of them (e.g. outlined-text export), `_extract` now returns `extractable: false` with an explicit reason instead of a hollow 0-bar/0-note success. Verified with a synthetic PDF that has staff lines but no digit text.

9. **Last-column duration** - `_infer_measure_rhythm` now takes the measure's actual barline position (`bar_end_x`) and uses `bar_end_x - last_column_x` for the final gap, instead of averaging the preceding gaps. Direct unit test pins the arithmetic.

10. **`_detect_barlines` re-parsing per staff** - `_vertical_segments(page)` is now computed once per page and passed in, instead of being recomputed inside `_detect_barlines` for every staff on that page.

**Also**: added `DELETE /api/scores/{id}/transcription`, which deletes only the `source='edited'` row (leaving `extracted` untouched) and returns whatever remains, 404 if nothing does. A second delete (or deleting when nothing was ever edited) is a clean no-op/404, not an error. Test covers edit -> delete -> GET-returns-extracted, plus the no-op and 404 paths.

**Verification**
- `pytest -q` inside a clean `python:3.12-slim` container against a copy of the real library: 31 passed.
- Rebuilt the full Docker image and ran it against the real library: scanned "To Zanarkand", transcribed with `(3,4)`, got 50 bars / 357 notes / 243 beats, tuning line `\tuning E4 B3 G3 D3 A2 D2` (Drop D, high string first), one rest bar preserved, and a new warning surfacing the two above-ceiling native-span frets. Exercised PUT (edit) -> GET -> DELETE -> GET (reverts to extracted) -> DELETE again (no-op, still extracted) -> DELETE on an unrelated score id (404). Confirmed the `time_signature` 422s on `(0,4)` and `(4,3)`.

PR #35, opened 2026-08-19
Decode tab rhythm and time signature from music-font glyphs (integrated into the production extractor)

sirged:
## Summary

Fermata's tab extractor reads fret numbers reliably (plain ASCII text with
coordinates) but guessed note *durations* from x-spacing between columns,
and couldn't read time signatures at all — because Finale (Maestro) and
Sibelius (Opus/OpusSpecial) draw noteheads, stems, flags, rests and
time-signature digits in a subsetted, re-encoded embedded music font.
Extracted "text" for those glyphs is meaningless remapped Unicode, not
digits or note names.

**This PR now integrates that glyph decoding directly into the production
extractor** (`server/fermata/tabextract.py`), rather than leaving it in a
standalone tool that duplicated staff/fret extraction. `glyph_rhythm.py`
moved to `server/fermata/` as a real product module; `tabextract.extract()`
uses it for rhythm and time signature whenever a tab staff is paired with a
standard-notation staff drawn in a font it can decode, falling back to the
existing x-spacing heuristic otherwise. Production's own staff/fret/chord
extraction and its already-fixed high-string-first tuning order are
untouched and stay authoritative — the glyph decoder only ever supplies
rhythm/time-signature/rest data.

The standalone `tools/tab_extract/extract_tab.py` and
`extract_tab_baseline.py` are deleted; their unique value (the glyph
decoding) is now in the product. `verify_tex.mjs` is kept — it parses
emitted alphaTex with the real alphaTab JS importer, which the Python test
suite can't do on its own, and is the actual regression check for
alphaTex's `{d}`/`{dd}` dotted-beat syntax. `fonttools` is now a required
runtime dependency (`server/pyproject.toml`), not an optional tools-only
extra.

## How the glyph decoding works

- **Finale/Maestro**: glyph *names* are stripped (post table format 3), but
  every Finale export sampled from the library embeds the exact same
  204-glyph subset, and a given glyph ID's outline coordinates are
  byte-identical across files — confirmed by direct coordinate diffing.
  Glyph ID is therefore a stable, verified lookup key.
- **Sibelius/Opus family**: subsets are per-file, but the post-table glyph
  *names* (PUA labels like `uniF0CF`) stay stable across files even though
  glyph IDs don't, so those are keyed by name instead.
- Every glyph→meaning mapping (notehead shapes, flag hook counts, rest
  types, digits, common/cut time) was established by rendering the real
  outline to a contact sheet and reading it by eye — not guessed — then
  cross-checked against occurrence counts and shape metrics.
- Stems and beams are vector line/rectangle primitives, not glyphs.
  Duration comes from combining notehead shape + stem presence + flag-glyph
  count at the stem's free end + stacked beam-rectangle count + trailing
  augmentation-dot glyphs.
- Tie detection is low-confidence and never used to merge durations — only
  surfaced as an honest warning.

## What changed in this integration pass

- **Rhythm source selection is per tab staff, not global**: each tab staff
  is paired with the nearest standard staff above it; if that pairing finds
  real glyph events, the whole staff's rhythm (and, per-page, the time
  signature) comes from glyph decoding. Otherwise it falls back to the
  original spacing heuristic. A single document can end up mixed (e.g. one
  page raster, one page vector) and the warnings say so honestly.
- **Warnings are now honest about what was actually used**: the old
  unconditional "durations are inferred from horizontal spacing... treat as
  low confidence" and "time signature not detected" warnings only appear
  when that's still true. When glyph decoding succeeded, they're replaced
  with warnings for what's genuinely still uncertain — tuplets aren't
  detected, tie detection is low confidence, and any tab columns that
  couldn't be matched to a decoded note are called out with a count.
- **Tuning order bug is not reintroduced**: the tool's `\tuning` emission
  was low-to-high (mirrored strings); production's fixed high-to-low-string
  emission is the only one wired in. Verified directly against alphaTab:
  the first note of this Drop D piece plays MIDI 64, not 38/40.

## Verification

`server/tests/test_tabextract.py` now asserts, against a real library fixture (a solo-guitar arrangement):
- the time signature (3/4) is read via glyph decoding with **no manual
  override**;
- dotted durations are present and the emitted alphaTex actually **parses**
  with alphaTab's real JS importer (this is the regression that blocked the
  original version of this PR — a duration code with a trailing dot like
  `:8.` looks plausible but alphaTab rejects it; the correct spelling is
  `:8 ...{d}`);
- the emitted `\tuning` line is high-string-first and the first parsed note
  is MIDI 64;
- the old fallback paths (raster PDF, notation-only PDF) still behave
  identically.

Ran the full suite (32 tests) in a clean container with the real library
mounted read-only, and separately built the production Docker image from
this branch, ran it against the real library, and transcribed that same library fixture through the actual HTTP API without passing a time signature.
See the PR comment below for verbatim output of both.

## Scope / known gaps (unchanged from before)

- Sibelius exports that embed Opus as CFF (Type1C) rather than TrueType
  aren't covered — falls back to the spacing heuristic gracefully, no
  crash, no wrong data.
- Tuplets aren't detected; a triplet's three notated eighths still report
  as plain eighths.
- Tie merging isn't implemented, only the low-confidence hint, surfaced as
  a warning.


sirged:
Addressed all 11 findings from the review, scoped to `server/tools/tab_extract/**`.

## Fixes

**1. CRITICAL - unparseable dotted notes.** `_fmt_beat` no longer emits a trailing dot on the duration code (`:8.`); dots are now the `{d}`/`{dd}` beat effect (`:8 3.4{d}`). Added `verify_tex.mjs`, a standing regression check that parses every generated `.tex` with the real alphaTab importer - this is what caught the original bug and now guards against it recurring.

**2. Time signature pairing/missing digits.** `decode_time_signature` now groups digit glyphs by x-adjacency into numerator/denominator runs (handles 12/8 correctly) and requires an unambiguous best x-overlap pairing before trusting it, else returns "not detected". Added `digit7`/`digit9` to `MAESTRO_GID_MAP` - confirmed by rendering the actual glyph outlines from real library files (`Moonlit Shadows (New World).pdf`, a 7/8 piece; `The Butterfly (New World).pdf`, 9/8) and eyeballing them, same method the rest of the table was built with. `digit0` is deliberately left unmapped - it never appears anywhere in the library's Maestro-subset pages, so there's no real outline to confirm a GID against.

**3+4. Beam detection / ties+stems starved.** `extract_stems_beams_curves` now groups each drawing path's line items into individual contours and measures each one's own perpendicular thickness (polygon area / long-axis length) instead of the whole path's y-bbox height - this correctly separates a two-level beam group into two beams instead of one wrong union, and stops discarding a real single beam just because it's steeply slanted (confirmed against `Classical-Guitar-Method-Vol1-2020.pdf` p92, a beam with a 16.7pt bbox height over an 85pt run that the old flat height<=14 gate always rejected - found 16 more real cases like it across the library). The old code's blanket `continue` after a beam-shape match also starved curve/stem detection for anything sharing that path; restructured so beams, stems, and curves (ties/slurs) are all found independently.

**5. Beam-ownership / max() bias.** `_beam_count_near` now restricts to the stem's free end (matching its own docstring and `_flag_count_near`'s existing logic) instead of the whole stem span, and duration decode now prefers the closest-matching stem among candidates instead of `max()` across all of them - a second voice's beam can no longer upgrade a note's duration.

**6. Font resource handling.** `MusicFont` is now keyed and tracked per xref (`load_music_fonts` returns `{family: [MusicFont, ...]}`), not one per family name - confirmed a real page with two differently-tagged `OpusSpecial` resources (`Easy-Christmas-Songs-for-Guitar-Vol1-4.pdf` p5/p31, though those specific resources are CFF and still out of scope per the existing `ext not in (ttf, otf)` gate). `extract_glyph_events` now tries each same-family candidate per glyph and also normalizes the span font name defensively (`.split("+")[-1]`).

**7. Duration matching tolerance / reference point.** `x_tol` tightened from `spacing*8.0` (~50-60pt) to `spacing*2.5`. Column x is now compared against an approximated column *center* (`col["x"] + avg_digit_w/2`, using the staff's own measured average digit width) instead of the digit's raw left edge vs. the notehead's center.

**8. Rests dropped / rest-only bars vanish.** `build_measure_beats` is now driven by the full note+rest glyph-decoded event list per measure instead of only pitched notes matched per tab column: rests emit as real `r` beats, and a rest-only bar (zero tab digits) is no longer skipped.

**9. Dead code.** Removed the unused `beats_per_measure` and the per-system time-signature propagation loop that filled a dict nothing read. Documented the resulting limitation honestly: one time signature is decoded per page (first system that states one) and used for the whole page's `\ts` header; a genuine mid-page meter change isn't modeled in the emitted header (later systems still decode their own correct rhythm regardless).

**10. Early-return page_no bug.** Both early-return branches in `process_page` now pass `page_no` to `_save_report`/`_render_png`.

**11. fontTools undeclared dependency.** Added `[project.optional-dependencies] tools = ["fonttools>=4.50"]` to `server/pyproject.toml`, documented in `glyph_rhythm.py`.

**Minor notes:** removed the unused `collections` import (re-added later, now genuinely used by a chord-clustering helper - see below); `OpusText` is now excluded from font loading entirely (it holds fingering/annotation text glyphs, not music symbols - loading it only inflated `unknown_glyphs` with irrelevant characters); `extract_glyph_events` now caches per page via a `WeakKeyDictionary` instead of re-parsing embedded fonts and re-walking texttrace on every call.

## One additional bug found during verification (not in the original list)

Matching each glyph-decoded pitched note independently against tab columns let a chord's 2nd/3rd/4th notehead (all sharing one x) each look for its own nearest *unused* column once the first member claimed the correct one - in a dense passage the nearest still-unused column could be the *next beat's* column, silently stealing it and corrupting that beat's duration. Confirmed via a per-measure quarter-note-sum sanity check against each page's own time signature. Fixed by clustering same-x glyph events into one beat before matching (`_cluster_pitched_events`/`_cluster_duration`) - this alone cut spurious "unmatched glyph note" counts by 60-100% across the sample files (e.g. To Zanarkand p0: 61 -> 0; Down By the River p0: 87 -> 26; 1 AM: 8 -> 0).

## Verification

**Parse-check (finding 1's regression test):** 12 pages across Finale (`To Zanarkand`, `Down By the River` x2, `1 AM`, `Hinata vs Neji`) and Sibelius/Opus (`Classical-Guitar-Method-Vol1-2020` x3, `Pachelbels-Canon-Werner-Guitar` x2, `RecuerdosdelaAlhambraTAB` x2) sources - **all 12 parse cleanly** with alphaTab's own `AlphaTexImporter` (`server/tools/tab_extract/verify_tex.mjs`). `Tarrega-Study-in-C-Guitar-Free.pdf` (the Sibelius standard-notation-only file with no tab staff) correctly produces no `.tex` at all - zero parse failures either way.

| file | bars | beats | notes |
|---|---|---|---|
| To Zanarkand p0 | 24 | 123 | 185 |
| Down By the River p0 | 20 | 187 | 222 |
| Down By the River p1 | 17 | 157 | 196 |
| 1 AM p0 | 17 | 113 | 103 |
| Hinata vs Neji p0 | 10 | 74 | 82 |
| Classical Guitar Method p83 | 18 | 111 | 108 |
| Classical Guitar Method p88 | 12 | 96 | 192 |
| Classical Guitar Method p93 | 9 | 36 | 36 |
| Pachelbels-Canon p2 | 12 | 120 | 120 |
| Pachelbels-Canon p3 | 14 | 132 | 173 |
| RecuerdosdelaAlhambraTAB p0 | 10 | 223 | 223 |
| RecuerdosdelaAlhambraTAB p1 | 12 | 258 | 258 |

**Visual re-verification:** rendered annotated overlays (decoded duration labels against the actual PDF engraving) for To Zanarkand p0 and Down By the River p0 (the dense 2-voice Baldur's Gate 3 piece used before). Both read correctly against the engraving after the beam/stem changes - beam groupings, the whole/dotted-half chords, and the sustained-bass-vs-16th-melody 2-voice passage all match what's actually printed.

**Cross-file scan (finding 3's regression cases):** across the full library, 16 real slanted-beam cases previously discarded by the old flat height gate are now correctly detected (not synthetic - confirmed on `Classical-Guitar-Method-Vol1-2020.pdf`, several `Final Fantasy` pieces, etc).

**Full-corpus regression check** (old vs. new code, same 120-PDF/472-page sample, both run to completion, zero errors either way):

| | old (pre-fix) | new (this PR) |
|---|---|---|
| extractable pages | 288 | 288 |
| total fret numbers extracted | 37,810 | 37,810 (identical) |
| total measures | 4,109 | 4,123 |
| rest beats emitted | n/a (always dropped) | 766 |
| unmatched columns (honest count) | ~0 (vacuous, tolerance never rejected) | 3,388 |
| unmatched glyph notes (honest count) | not tracked | 1,409 |

Fret-number extraction is untouched (identical count, as expected - none of these findings touch that code path). The unmatched counts are **not** tuned toward zero - they're now honestly non-trivial, which is the intended effect of findings 7/8 (a vacuously-wide tolerance previously always found *some* match). Spot-checking the largest single-page unmatched-column count (`Classical-Guitar-Method-Vol1-2020.pdf` p83, 71/108 unmatched) against the rendered page confirmed it's a real, pre-existing limitation of this decode approach - an arpeggio exercise ("Scarborough Fair", 6 tab plucks per measure realizing a chord) where the standard-notation staff genuinely only writes the top melodic line, not one notehead per pluck. Similarly, the Down By the River 2-voice piece's remaining unmatched glyph notes reflect genuine overlapping-voice content (a sustained melody note and a differently-valued bass note at the same onset x) that a single-voice-per-beat model can't fully reconcile - documented in `build_measure_beats`'s docstring rather than papered over.

sirged:
## Integration verification

### 1. Test suite, clean container, real library mounted read-only at `/testlib`

Built a throwaway python+node test image (production's own image is
python-only, but the alphaTex JS-parse regression check needs the real
alphaTab importer) and ran:

```
docker run --rm -v "<library>:/testlib:ro" -e FERMATA_TEST_LIBRARY=/testlib fermata-glyph-test pytest tests/ -v
```

Verbatim output:

```
============================= test session starts ==============================
platform linux -- Python 3.12.14, pytest-9.1.1, pluggy-1.6.0 -- /usr/local/bin/python3.12
cachedir: .pytest_cache
rootdir: /app/server
configfile: pyproject.toml
plugins: anyio-4.14.2
collecting ... collected 32 items

tests/test_tabextract.py::test_finale_tab_pdf_extracts_notes_and_bars PASSED [  3%]
tests/test_tabextract.py::test_finale_tab_pdf_analyze PASSED             [  6%]
tests/test_tabextract.py::test_glyph_decoded_time_signature_and_dots_without_override PASSED [  9%]
tests/test_tabextract.py::test_glyph_decoded_alphatex_parses_with_dotted_beats PASSED [ 12%]
tests/test_tabextract.py::test_notation_only_pdf_has_no_tab_staves PASSED [ 15%]
tests/test_tabextract.py::test_raster_pdf_is_not_extractable PASSED      [ 18%]
tests/test_tabextract.py::test_malformed_pdf_never_raises PASSED         [ 21%]
tests/test_tabextract.py::test_missing_pdf_never_raises PASSED           [ 25%]
tests/test_tabextract.py::test_tuning_emitted_high_string_first_for_alphatex_binding PASSED [ 28%]
tests/test_tabextract.py::test_measure_quarter_length_scales_by_denominator PASSED [ 31%]
tests/test_tabextract.py::test_six_eight_and_three_four_extract_identically PASSED [ 34%]
tests/test_tabextract.py::test_escape_tex_string_handles_quotes_and_backslashes PASSED [ 37%]
tests/test_tabextract.py::test_hostile_title_is_escaped_in_alphatex PASSED [ 40%]
tests/test_tabextract.py::test_rest_only_bar_is_emitted_not_skipped PASSED [ 43%]
tests/test_tabextract.py::test_last_column_duration_uses_barline_not_average_gap PASSED [ 46%]
tests/test_tabextract.py::test_tab_staff_count_is_summed_across_pages PASSED [ 50%]
tests/test_tabextract.py::test_digit_merge_rejects_impossible_frets PASSED [ 53%]
tests/test_tabextract.py::test_digit_merge_flags_suspicious_native_two_digit_span PASSED [ 56%]
tests/test_tabextract.py::test_tab_staff_found_but_no_digits_is_not_extractable PASSED [ 59%]
tests/test_transcription_api.py::test_edited_transcription_survives_re_extraction PASSED [ 62%]
tests/test_transcription_api.py::test_get_transcription_404_when_none_exists PASSED [ 65%]
tests/test_transcription_api.py::test_has_transcription_flag_is_batched_and_accurate PASSED [ 68%]
tests/test_transcription_api.py::test_transcribe_rejects_non_extractable_pdf PASSED [ 71%]
tests/test_transcription_api.py::test_transcription_analysis_endpoint PASSED [ 75%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature0] PASSED [ 78%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature1] PASSED [ 81%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature2] PASSED [ 84%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature3] PASSED [ 87%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature4] PASSED [ 90%]
tests/test_transcription_api.py::test_transcribe_accepts_valid_time_signature PASSED [ 93%]
tests/test_transcription_api.py::test_delete_transcription_reverts_to_extracted PASSED [ 96%]
tests/test_transcription_api.py::test_delete_transcription_404_when_none_exists PASSED [100%]

============================== 32 passed in 3.60s ==============================
```

### 2. End-to-end proof: production image, real API, no manual time signature

Built the production `Dockerfile` from this branch, ran it against the real
library (mounted read-only), triggered a scan (298 files, 0 errors), then:

```
POST /api/scores/216/transcribe   (body: {} - no time_signature)
```

Response (trimmed to the relevant fields):

```json
{
  "bars": 50,
  "beats": 243,
  "notes": 357,
  "tempo": 88,
  "tuning": ["D2","A2","D3","G3","B3","E4"],
  "tuning_label": "Drop D",
  "time_signature": [3, 4],
  "time_signature_source": "glyph-decoded",
  "warnings": [
    "tuplets (triplets and similar) are not detected - a note written inside a tuplet will show its plain written duration rather than the shortened tuplet duration",
    "tie detection is low confidence - some tied notes may show up as separately re-struck notes instead of one held note",
    "1 fret number(s) could not be matched to a note in the engraved notation and got an estimated duration instead - treat those specific notes as low confidence",
    "2 note(s) read from the engraved notation had no matching fret number and were dropped from the tab",
    "19 digit token(s) near a tab staff could not be assigned to a string",
    "2 fret number(s) above 24 were read directly from the PDF's own text (not from a merge) - likely two adjacent notes rendered as one text span in the source - treat those frets as low confidence"
  ]
}
```

The old "note durations are inferred from horizontal spacing... treat as
low confidence" and "time signature not detected" warnings are gone, as
expected - both claims are now false for this file.

Stored alphaTex header (`GET /api/scores/216/transcription`):

```
\title "To Zanarkand (Final Fantasy X)"
\tempo 88
\ts 3 4
\tuning E4 B3 G3 D3 A2 D2
```

`\tuning` is high-string-first as required (`E4 B3 G3 D3 A2 D2`, not the
tool's old mirrored `D2 A2 D3 G3 B3 E4`).

### 3. Parsed the stored transcription with alphaTab's real importer

Per the verification method noted in `verify_tex.mjs`, ran the *stored*
transcription content (fetched from the running container's API, not
regenerated) through the actual importer at
`web/node_modules/@coderline/alphatab/dist/alphaTab.mjs`:

```
node server/tools/tab_extract/verify_tex.mjs <alphaTab.mjs> zanarkand_stored.tex
{"ok":true,"bars":50,"beats":243,"notes":357,"dottedBeats":30,"firstNoteMidi":64}
```

- Parses cleanly (the `{d}`/`{dd}` dotted-beat regression is fixed).
- `dottedBeats: 30` - dotted durations are really present in the parsed
  score, not just in the source text.
- `firstNoteMidi: 64` - the first note (`0.1`) sounds high E, confirming
  the tuning line is emitted in the correct (high-string-first) order for
  this Drop D piece, not the mirrored 38/40 a low-to-high emission would
  produce.

sirged:
## Hardened the glyph rhythm decoder against silent mis-decoding

Pushed as `9b4d807`. The governing rule throughout: **an honest fallback with a warning beats confidently-wrong output.** Where a signal can't be trusted it now degrades to the spacing heuristic and says why, instead of emitting plausible-looking wrong rhythm at "high confidence".

Every threshold below was derived from measurement on the library first, and the numbers reported here are the real ones — no thresholds were tuned until warnings read zero.

---

### Correctness findings

**1 (HIGHEST) — Maestro GID map gated only on the font family name.** Now fingerprinted. At font-load time the raw `glyf` bytes of each mapped GID are hashed and compared against `MAESTRO_GLYF_DIGESTS`; a non-matching font is treated as unrecognised so the spacing fallback engages, with a warning naming the situation.

The module's assumption was verified before being relied on: across **548 embedded Maestro resources in 274 library files**, 36 mapped GIDs are in use and **zero have more than one distinct outline**. That's what makes GID a legitimate key for this family. Subsets keep all 204 slots but fill only 6–22 of the mapped ones (median 14), so absent glyphs are not evidence either way and the check requires 4 present-and-matching (`MAESTRO_FINGERPRINT_MIN_GLYPHS`).

Both directions proven:

| input | result |
|---|---|
| the library's own embedded Maestro | accepted, 11 outlines match |
| one mapped glyph's outline nudged | rejected |
| glyph order permuted (a different subsetting path) | rejected |
| an unrelated TrueType font renamed Maestro | rejected, 35 outlines differ |
| unmodified fontTools round-trip | accepted (check isn't brittle) |

End-to-end with the digest table poisoned, the reference file degrades honestly — fret extraction untouched (357 notes), rhythm `low`, provenance `{'spacing': 10}`, and **no** glyph-decoded time signature, because the digit GIDs can no longer be trusted either.

**Recalibrating:** `python server/tools/tab_extract/maestro_fingerprint.py <library-root>` re-derives the table and prints it ready to paste. Read its "GIDs with MORE THAN ONE distinct outline" section first — a GID listed there must be **dropped** from `MAESTRO_GID_MAP`, not "updated" to whichever file was scanned last, since disagreement means the subset is no longer stable. It also prints the present-GIDs-per-resource floor so `MAESTRO_FINGERPRINT_MIN_GLYPHS` stays below it.

**2 (HIGH) — `_pair_standard_staff` could pair across systems.** Staves are grouped into systems by vertical gap, then paired only within one, with a spacing-relative distance ceiling, an x-overlap requirement, and ambiguity returning nothing. Measured to set the constants: the notation-to-tab gap is 3.3–7.3 tab-staff spacings (n=254, median 5.4) and x-overlap is 100% of the narrower staff on all 254 pairs, so a 12-spacing ceiling and 0.6 overlap have wide margins either side.

**3 (HIGH) — the decoder's honesty stats were discarded.** Propagated and acted on. `unknown_ratio` above 0.35 falls back for that staff; above 0.10, or any unrecognised glyph sitting where a flag attaches, downgrades to `medium` and names the offending GIDs. Thresholds sit far above what the calibrated vocabulary produces: the ratio is 0.0 at the 99th percentile of notation staves and peaks at 0.065.

This immediately surfaced a real calibration gap rather than a false alarm: **Bloom and Breathe (New World).pdf** uses Maestro **GID 120** (6 occurrences, 9.7×20.5pt) which isn't in the map — that file is now reported `medium` instead of `high`. Worth calibrating in a follow-up.

**4 (MEDIUM) — time signature taken from the first staff and applied document-wide.** Now a per-system timeline, carried forward and emitted at the bar where the meter changes. **15 library files change meter mid-score** and were all silently measured against their opening meter. Concrete case: *Weeping Dawn (Baldur's Gate 3)* is 4/4 → 12/16 → 4/4; the emitted alphaTex now carries `\ts 12 16` at bar 24 and `\ts 4 4` at bar 31, and alphaTab parses it. The decoder's diagnostic reasons are also surfaced now (49 files gain e.g. `time signature: only 0 time-signature digit glyph(s) found` instead of only a generic "assumed 4/4").

**5 (MEDIUM) — absolute point tolerances.** All ~20 distances are now named constants in staff spaces (`_SP`, applied via `_Tol`), scaled by the staff being decoded. They're the old values converted at the library's measured 5.125pt reference spacing, so behaviour on the sampled library is unchanged. Tested at 2.0/5.125/12.0pt spacings.

**6 (MEDIUM) — provenance duplicated in three hand-synchronised places.** Each tab staff resolves to one `_RhythmSource` (`PROV_GLYPHS` / `PROV_GLYPHS_DEGRADED` / `PROV_SPACING`); `_rhythm_report` derives the document's warnings *and* confidence from the collected provenances in one function. Provenance is also reported on `ExtractionResult.rhythm_provenance`.

---

### Second-round findings

**A1 — glyph-decoded signatures bypassed validation.** Fixed, and this was worse in practice than reported. Digit runs no longer merge across rows (which turned 1/16 into "116"), and any signature must be a usable meter or it's reported as not detected. `api.py` now imports the rule instead of keeping a second copy.

It wasn't only an unrenderable stored `\ts`: an unvalidated denominator of **0** out of the plain-text digit scan hit `_measure_quarter_length` and raised `ZeroDivisionError`, which `extract()`'s blanket handler turned into `extractable: false` for a file whose fret digits read fine. **Easy-Christmas-Songs-for-Guitar-Vol1-4.pdf** was affected on `main`; it now extracts (351 bars, 1826 notes).

**A2 — `fontTools` imported at module scope.** Now lazy and guarded; a missing install degrades to the spacing heuristic instead of taking down the whole server including `/api/health`.

**A3 — `getGlyphOrder()` outside the try/except.** Font loading, glyph-order access and the fingerprint check are all inside the guard; a corrupt embedded font degrades to fallback rather than failing an extraction whose fret digits never needed the font.

**B1 — every rest glyph became its own beat.** Rests are clustered by onset and dropped where a pitched onset already sits (merge window 0.8 staff spacings — well under the 1.25 spacings between consecutive sixteenths in this engraving); simultaneous rests in two voices collapse to one beat. **176 phantom rest beats removed across 45 files, with bars and notes unchanged** (rest beats 323 → 147).

**B2 — two tab staves sharing one notation staff.** Fixed structurally by finding 2's exclusive pairing: a notation staff is read by at most one tab staff, and the other is honestly reported as having none.

**B3 — no gates on the filled-rect beam branch.** Added max-width (half the staff) and min-thickness gates. Reported honestly: I could not reproduce this on this library — no exporter here draws staff lines as wide dark filled rects (0 occurrences across 108 pages), so it's a latent hazard rather than an active bug. The gate is cheap and correct regardless.

**B4 — best-stem selection skipped the note's own plain stem.** Now picks the closest stem instead of the highest-scoring one. Verified against the engraving rather than assumed: I rendered the three affected noteheads in the reference file and each is a plain up-stem with **no beam or flag** — genuine quarters that `main` emits as eighths by stealing the lower voice's beam.

**B5 — beam matched on bbox centre.** `Beam` now carries centreline endpoints and `beam_y_at()` interpolates at the stem's x. No change to beam counts on the reference file (identical per-staff totals), so this is pure upside for the slanted groups it was written for.

**B6 — barlines passed the stem gate.** Confirmed: **2089 of 16052** accepted "stems" were exact-staff-height verticals. The fix needed care — my first attempt ("spans most of the staff") also swallowed the real down-stem of a note above the staff and turned a beamed eighth into a quarter. Measurement settled it: real barlines land within **0.03pt** of both outer staff lines, while that stem misses by 0.88pt and 1.30pt, so the test is now tight and two-sided (0.1 staff spaces).

**B7 — Opus `uniF0CE` never counted as a flag.** Added to `FLAG_HOOKS`, so the code now does what its comment claimed. This path is genuinely exercised — 10 Opus-family library files have tab staves — and *Spanish-Romance* shows 7 notes moving from quarter to eighth as predicted.

**B8 — `ext` accepted `"otf"`.** Fixed, with a correction to the premise: **pymupdf reports CFF embeds as `"cff"`, not `"otf"`** (measured: 548 Maestro/ttf, 36 Opus-family/ttf, 31 Opus-family/cff, 0 otf), and `"cff"` was already excluded — so the described mis-decode path does not occur here. Gated on the loaded font actually having `glyf` outlines anyway, which is the real semantic test.

---

### Efficiency findings (7–12, all done)

Cached `get_drawings()` per page and embedded fonts per document (stored on the document, since pymupdf's `Document` can't be weak-referenced and `id()` keying would hand a later document the previous one's fonts); tab census before any glyph work; bisect-bounded measure and column lookups; glyphs partitioned by category once per staff; oversized contours short-circuit the beam test's all-pairs scan.

| file | pages | before | after |
|---|---|---|---|
| Classical-Guitar-Method-Vol1-2020 | 102 | 391.6 ms | **295.5 ms** |
| Theme of Mitsuha | 3 | 163.7 ms | **60.5 ms** |
| To Zanarkand | 2 | 88.9 ms | **35.4 ms** |
| Tarrega-Study-in-C (notation-only) | 1 | 8.3 ms | **5.9 ms** |
| whole library, 292 files | — | 26.1 s | **11.6 s** |

---

### Verification

**No regression on the reference file** — `To Zanarkand (Final Fantasy X).pdf`, parsed with the real importer at `web/node_modules/@coderline/alphatab/dist/alphaTab.mjs`:

```
ts=(3,4) source=glyph-decoded  bars=50  beats=243  notes=357
alphaTab: ok=true bars=50 beats=243 notes=357 dottedBeats=30 firstNoteMidi=64
provenance={'glyphs': 10}   rhythm=high
```

The emitted alphaTex differs from `main` in exactly two places, both fixes — `:2 (...){dd}` → `{d}` in bars 4 and 24. A double-dotted half is 3.5 quarters, which cannot fit in a 3/4 bar; it came from one voice's dot being claimed by a neighbouring voice's note.

**Broad scan, whole library (292 files):**

| | main | this branch |
|---|---|---|
| extractable | 287 | **288** (crash fixed) |
| crashes | 0 (1 masked as `extractable: false`) | **0** |
| tab staves glyph-decoded | — | 2561 |
| tab staves degraded / spacing | — | 1 / 130 |
| files all-glyph `high` | 278 | 277 |
| files `medium` (degraded) | 0 | 1 |
| files `low` (all spacing) | 9 | 10 |
| time signature glyph-decoded | 229 | 229 |
| bars summing to their meter | 14.6% | **17.1%** |

The whole coverage delta is accounted for and is not a loss: *Bloom and Breathe* moves `high` → `medium` (the real GID-120 gap above), and *Easy-Christmas-Songs* moves not-extractable → `low` (the crash fix; it embeds Opus as CFF, so spacing is the honest answer). **No file lost glyph decoding to the stricter gates.** The 130 spacing fallbacks are 10 files that were already falling back on `main` — they now say why (CFF-embedded Opus, or no readable note glyphs).

**Honest note on the bar-fullness metric.** I added it as an objective check rather than trusting unchanged counts, and it says 17.1%, not ~100%. The dominant residual error is **two-voice writing collapsed into one voice** — a documented modelling gap this change doesn't address (a bass quarter plus treble eighths at one onset becomes one beat with one duration, so the bar over-sums). My fixes reduce total absolute bar error from 3021.6 to 2918.1 quarters but cannot close it without voice separation. I've called this out in the module docstring; it's the obvious next piece of work and I'd suggest a follow-up issue.

**Tests** — 80 passed, in a clean `python:3.12-slim` container with the library mounted at `/testlib` (not `/lib`). New: 25 in `test_glyph_rhythm.py` (fingerprint accept/reject/evidence-floor, barline-vs-stem, beam interpolation and gates, stem selection, Opus flag glyph, dot assignment, staff-space scaling, signature validation, missing-fontTools) and 22 added to `test_tabextract.py` (system pairing, exclusivity, ambiguity, rest dedupe, the stats-driven downgrade, meter timeline, end-to-end fingerprint rejection *and* acceptance).

```
=== environment ===
Python 3.12.14
node: v20.19.2
pymupdf 1.28.2
fonttools 4.63.0
=== library mount ===
Classical
Favorites
Holiday
Method Books
Patreon
=== pytest ===
============================= test session starts ==============================
platform linux -- Python 3.12.14, pytest-9.1.1, pluggy-1.6.0 -- /usr/local/bin/python
rootdir: /work/server
configfile: pyproject.toml
plugins: anyio-4.14.2
collecting ... collected 80 items

tests/test_glyph_rhythm.py::test_fingerprint_accepts_the_calibrated_library_font PASSED [  1%]
tests/test_glyph_rhythm.py::test_fingerprint_rejects_altered_glyph_outlines PASSED [  2%]
tests/test_glyph_rhythm.py::test_fingerprint_rejects_when_every_digest_disagrees PASSED [  3%]
tests/test_glyph_rhythm.py::test_fingerprint_needs_enough_evidence_to_bless_a_font PASSED [  5%]
tests/test_glyph_rhythm.py::test_barline_is_not_mistaken_for_a_stem PASSED [  6%]
tests/test_glyph_rhythm.py::test_beam_carries_interpolated_end_ys_not_a_bbox_centre PASSED [  7%]
tests/test_glyph_rhythm.py::test_slanted_beam_is_counted_at_both_ends_of_the_group PASSED [  8%]
tests/test_glyph_rhythm.py::test_staff_line_shaped_rect_is_not_a_beam PASSED [ 10%]
tests/test_glyph_rhythm.py::test_beam_detection_short_circuits_oversized_contours PASSED [ 11%]
tests/test_glyph_rhythm.py::test_note_keeps_its_own_plain_stem_against_a_neighbours_beam PASSED [ 12%]
tests/test_glyph_rhythm.py::test_stem_selection_prefers_the_closest_candidate PASSED [ 13%]
tests/test_glyph_rhythm.py::test_opus_shared_flag_rest_glyph_counts_as_a_flag PASSED [ 15%]
tests/test_glyph_rhythm.py::test_one_dot_belongs_to_exactly_one_note PASSED [ 16%]
tests/test_glyph_rhythm.py::test_double_dot_is_still_read_when_two_dots_really_are_there PASSED [ 17%]
tests/test_glyph_rhythm.py::test_geometry_decodes_the_same_at_any_staff_size[2.0] PASSED [ 18%]
tests/test_glyph_rhythm.py::test_geometry_decodes_the_same_at_any_staff_size[5.125] PASSED [ 20%]
tests/test_glyph_rhythm.py::test_geometry_decodes_the_same_at_any_staff_size[12.0] PASSED [ 21%]
tests/test_glyph_rhythm.py::test_tolerances_scale_linearly_with_spacing PASSED [ 22%]
tests/test_glyph_rhythm.py::test_tolerances_fall_back_to_the_reference_spacing_when_unknown PASSED [ 23%]
tests/test_glyph_rhythm.py::test_time_signature_validity_matches_the_api_rule PASSED [ 25%]
tests/test_glyph_rhythm.py::test_digit_clusters_do_not_merge_vertically_stacked_digits PASSED [ 26%]
tests/test_glyph_rhythm.py::test_unusable_signature_is_reported_as_not_detected PASSED [ 27%]
tests/test_glyph_rhythm.py::test_valid_stacked_signature_is_still_read PASSED [ 28%]
tests/test_glyph_rhythm.py::test_missing_fonttools_degrades_instead_of_crashing PASSED [ 30%]
tests/test_glyph_rhythm.py::test_non_truetype_embedding_is_refused_by_flavour PASSED [ 31%]
tests/test_tabextract.py::test_finale_tab_pdf_extracts_notes_and_bars PASSED [ 32%]
tests/test_tabextract.py::test_finale_tab_pdf_analyze PASSED             [ 33%]
tests/test_tabextract.py::test_glyph_decoded_time_signature_and_dots_without_override PASSED [ 35%]
tests/test_tabextract.py::test_glyph_decoded_alphatex_parses_with_dotted_beats PASSED [ 36%]
tests/test_tabextract.py::test_notation_only_pdf_has_no_tab_staves PASSED [ 37%]
tests/test_tabextract.py::test_raster_pdf_is_not_extractable PASSED      [ 38%]
tests/test_tabextract.py::test_malformed_pdf_never_raises PASSED         [ 40%]
tests/test_tabextract.py::test_missing_pdf_never_raises PASSED           [ 41%]
tests/test_tabextract.py::test_tuning_emitted_high_string_first_for_alphatex_binding PASSED [ 42%]
tests/test_tabextract.py::test_measure_quarter_length_scales_by_denominator PASSED [ 43%]
tests/test_tabextract.py::test_six_eight_and_three_four_extract_identically PASSED [ 45%]
tests/test_tabextract.py::test_escape_tex_string_handles_quotes_and_backslashes PASSED [ 46%]
tests/test_tabextract.py::test_hostile_title_is_escaped_in_alphatex PASSED [ 47%]
tests/test_tabextract.py::test_rest_only_bar_is_emitted_not_skipped PASSED [ 48%]
tests/test_tabextract.py::test_last_column_duration_uses_barline_not_average_gap PASSED [ 50%]
tests/test_tabextract.py::test_tab_staff_count_is_summed_across_pages PASSED [ 51%]
tests/test_tabextract.py::test_digit_merge_rejects_impossible_frets PASSED [ 52%]
tests/test_tabextract.py::test_digit_merge_flags_suspicious_native_two_digit_span PASSED [ 53%]
tests/test_tabextract.py::test_tab_staff_found_but_no_digits_is_not_extractable PASSED [ 55%]
tests/test_tabextract.py::test_tab_staff_pairs_with_the_notation_staff_in_its_own_system PASSED [ 56%]
tests/test_tabextract.py::test_tab_staff_does_not_pair_across_systems PASSED [ 57%]
tests/test_tabextract.py::test_tab_staff_does_not_pair_with_a_different_instruments_column PASSED [ 58%]
tests/test_tabextract.py::test_one_notation_staff_is_read_by_only_one_tab_staff PASSED [ 60%]
tests/test_tabextract.py::test_ambiguous_pairing_returns_no_notation_staff PASSED [ 61%]
tests/test_tabextract.py::test_second_voice_rest_under_a_note_is_not_an_extra_beat PASSED [ 62%]
tests/test_tabextract.py::test_simultaneous_rests_in_two_voices_collapse_to_one_beat PASSED [ 63%]
tests/test_tabextract.py::test_a_genuinely_separate_rest_is_still_its_own_beat PASSED [ 65%]
tests/test_tabextract.py::test_clean_decode_is_reported_as_glyph_sourced PASSED [ 66%]
tests/test_tabextract.py::test_unknown_glyphs_at_flag_positions_downgrade_confidence PASSED [ 67%]
tests/test_tabextract.py::test_mostly_unrecognised_vocabulary_falls_back_entirely PASSED [ 68%]
tests/test_tabextract.py::test_high_unknown_ratio_alone_downgrades PASSED [ 70%]
tests/test_tabextract.py::test_no_decoded_events_falls_back_with_the_font_reason PASSED [ 71%]
tests/test_tabextract.py::test_unpaired_tab_staff_falls_back_without_decoding PASSED [ 72%]
tests/test_tabextract.py::test_rhythm_report_downgrades_on_degraded_staves PASSED [ 73%]
tests/test_tabextract.py::test_rhythm_report_is_the_single_source_of_warnings_and_confidence PASSED [ 75%]
tests/test_tabextract.py::test_emitted_time_signature_is_always_usable PASSED [ 76%]
tests/test_tabextract.py::test_unusable_manual_override_is_ignored_not_emitted PASSED [ 77%]
tests/test_tabextract.py::test_zero_denominator_signature_never_reaches_the_measure_budget PASSED [ 78%]
tests/test_tabextract.py::test_mid_score_meter_change_is_carried_into_the_transcription PASSED [ 80%]
tests/test_tabextract.py::test_ts_timeline_lookup_uses_the_last_meter_printed_before_a_position PASSED [ 81%]
tests/test_tabextract.py::test_unrecognised_maestro_falls_back_honestly PASSED [ 82%]
tests/test_tabextract.py::test_calibrated_maestro_still_decodes PASSED   [ 83%]
tests/test_transcription_api.py::test_edited_transcription_survives_re_extraction PASSED [ 85%]
tests/test_transcription_api.py::test_get_transcription_404_when_none_exists PASSED [ 86%]
tests/test_transcription_api.py::test_has_transcription_flag_is_batched_and_accurate PASSED [ 87%]
tests/test_transcription_api.py::test_transcribe_rejects_non_extractable_pdf PASSED [ 88%]
tests/test_transcription_api.py::test_transcription_analysis_endpoint PASSED [ 90%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature0] PASSED [ 91%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature1] PASSED [ 92%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature2] PASSED [ 93%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature3] PASSED [ 95%]
tests/test_transcription_api.py::test_transcribe_rejects_invalid_time_signature[time_signature4] PASSED [ 96%]
tests/test_transcription_api.py::test_transcribe_accepts_valid_time_signature PASSED [ 97%]
tests/test_transcription_api.py::test_delete_transcription_reverts_to_extracted PASSED [ 98%]
tests/test_transcription_api.py::test_delete_transcription_404_when_none_exists PASSED [100%]

============================== 80 passed in 3.30s ==============================
```

### Suggested follow-ups

1. **Two-voice separation.** The dominant remaining rhythm error, and the thing standing between 17% and a high bar-fullness figure.
2. **Calibrate Maestro GID 120** (and re-check `_detect_barlines`, which uses the loose "spans most of the staff" test that B6 showed also matches real stems — it affects measure bounds, not durations, so I left it alone here).

PR #44, opened 2026-08-20
Separate voices when transcribing polyphonic music

### LINKED ISSUE #40: Separate voices when transcribing polyphonic guitar music
Transcription currently flattens everything on a staff into a single voice. For classical and fingerstyle guitar — which is polyphonic by nature, a melody sounding over an independent bass line — that means the bar contains the sum of both voices instead of the two of them running together.

Measured on a representative arrangement: 37 of 50 bars hold more than their time signature allows, typically about double, and only 24% of bars add up exactly. Each individual note's duration is decoded correctly from the engraving; it is the assembly that is wrong. Because playback follows the bar rather than the individual note, timing drifts through most of the score.

This is now reported honestly — the bar count appears in the transcription warnings and caps the rhythm confidence — but reporting it is not fixing it. Voice separation is the single largest remaining gap between a transcription and a score you could actually practise from.

**What it needs**

- Decide which voice each note belongs to. Stem direction is the primary signal in engraved notation (upper voice stems up, lower voice stems down), and the glyph decoder already recovers stems, so the information is largely in hand.
- Emit multiple voices rather than one sequence. alphaTex supports several voices per bar, so the target format is not the obstacle.
- Reconcile voices against the tab staff, where a chord and two independent voices can look similar.
- Verify against bar arithmetic rather than appearance: the share of bars that add up exactly is the metric, and it should move decisively.

**Why it is worth doing before further rhythm refinement**

Tuplets, ties and the remaining calibration gaps are each worth a few percent. This one accounts for most of the residual error on its own, and until it is addressed a transcription cannot be trusted for playback of the repertoire this project is aimed at.

sirged:
Closes #40.

Classical and fingerstyle guitar is polyphonic — a melody sounds *over* an independent bass line. The transcriber assembled every note on a staff into one sequence of beats, so a bar held the **sum** of its voices instead of its voices running concurrently: typically about double its meter, with every individual duration decoded correctly. Playback follows the bar, so timing drifted through most of a score.

## What changed

Each notehead is grouped with the others hanging off **its own engraved stem** (a chord is one beat), and where two stems genuinely sound at the same onset the bar is split into concurrent voices by **stem direction** — the signal engravers use for exactly this. Each voice is then filled out with rests so it accounts for the bar on its own, which is what actually makes the arithmetic work.

**Stem direction is deliberately not used on its own.** Single-voice writing also flips stems with pitch around the middle line, so splitting wherever direction changes would shred a melody crossing it into two voices at every crossing. **Simultaneity** is what distinguishes the two cases, because one voice never has two stems at one onset.

Three engraving details had to be handled for that to hold up:

- **A chord's inner noteheads sit part way *along* the shared stem**, far outside the tight end-window the duration pass uses (that window is what stops a neighbouring voice's stem being read for flags). They found no stem at all, so each became its own beat — and, at a shared onset, a phantom second voice. They now attach to the stem threaded through them, and a chord's duration is recomposed from all its members, since only the notehead at the stem's end can see the flag or beam.
- **A notehead whose stem is not found at all joins whatever else sounds at its onset.** Stems are what two-voice writing is notated *with*, so an onset carrying no stem evidence is a chord, not two voices. Some scores in the library engrave chords whose stem the vector pass cannot see; splitting those by pitch invented a voice out of one chord and left both halves short.
- **Two simultaneous notes in different voices are two tab digits on different strings — which is also exactly what a chord looks like.** The split cannot be read from the tab at all, so it is read from the notation: noteheads ordered by *pitch* against digits ordered by *string* (string 1 being the highest-pitched). That puts each voice's note on the string the engraver wrote it on whether the onset is one chord or two voices.

A whole note never takes a stem in any notation, so it is placed by where it sits relative to the voices that do have one. Half notes now have their stem looked up for direction only (never for a flag they cannot carry) — a lower voice in this repertoire is very often written in half notes and had no voice signal at all before.

## alphaTex syntax

Concurrent voices are separated by `\voice` **inside** the bar, which requires `\voicemode barwise` in the header. alphaTex's default (`staffwise`) reads `\voice` as "start the whole staff again for the next voice", so everything after the first separator would have become bar 1 onwards of voice 2. Verified against the installed renderer (alphaTab 1.8.4) rather than assumed. The directive is only emitted when a score actually has a multi-voice bar.

## Results

Bar arithmetic, measured through the emitted transcription. `exact` = every emitted voice sums exactly to its meter; `error` = total |voice quarters − budget|.

| File | exact before | exact after | error before | error after |
|---|---|---|---|---|
| Score A | 12/50 (24.0%) | **46/50 (92.0%)** | 72.50q | **6.00q** |
| Score B | 4/35 (11.4%) | **28/35 (80.0%)** | 51.50q | **15.50q** |
| Score C | 8/69 (11.6%) | **25/69 (36.2%)** | 178.50q | **74.50q** |
| Score D | 3/20 (15.0%) | **7/20 (35.0%)** | 32.00q | **15.50q** |
| Score E | 7/17 (41.2%) | **12/17 (70.6%)** | 32.00q | **8.00q** |
| Score F | 1/27 (3.7%) | **17/27 (63.0%)** | 75.50q | **15.75q** |
| Score G | 13/39 (33.3%) | **21/39 (53.8%)** | 30.75q | **13.50q** |

Whole library — 297 files, **291 extractable and 0 crashes, both unchanged**:

| | before | after |
|---|---|---|
| bars summing exactly | 1814 / 10689 (17.0%) | **7916 / 10689 (74.1%)** |
| overfull bars | 8017 | **2177** |
| total absolute error | 18707.06q | **3855.44q** |
| bars emitted as two voices | 0 | 7793 |

All 287 emitted transcriptions were re-parsed with the installed renderer: every one parses, and the renderer reports 7726 bars carrying more than one sounding voice — i.e. the `\voice` separators really land their beats in a second concurrent voice rather than merely parsing.

## What this does not fix

Tuplets, tie detection and time-signature calibration are unchanged and out of scope. Two of the residuals are worth naming because they now dominate:

- **Score C** is the weakest file above, and its residual is almost entirely *not* a voice problem: 53 of its overfull voices are over by exactly 1.0 quarter, which is a whole note measured against a 3/4 budget. The score prints 4/4, then 5/4, then 4/4; the decoder reads a change to 3/4 and misses the return. That is a time-signature bug.
- **Score A bar 10** is one of its 3 remaining overfull bars: the melody's third note is a flagged eighth in the score and decodes as a quarter, so that voice holds 3.5 quarters. A missed flag, not a missed voice.

Because a voice that is silent for part of a bar gets padded, a bar can no longer come out *underfull* in a multi-voice bar — so the honest comparison is the overfull count and the overfull error, and those move the same way (over-error 436.25q → 116.25q on the seven files above). The total quarters of rest that were deduced from the meter rather than read from a printed rest is reported as its own warning, so padded silence can be told from decoded silence.

`_overfull_bars` and the bars-that-do-not-add-up warning both stay, now reporting a much smaller number, and the warning names the causes that remain rather than blaming merged voices for all of them.

## Verification

- **Rendered and compared by eye.** Page 1 of Score A against the emitted transcription rendered by alphaTab: the melody lines up with the top voice on the high strings and the accompaniment with the lower voice on the low strings, bar for bar, including the bars where the two share an onset and their digits had to be split between the voices.
- **Tests** for the assignment rules — stem-up/stem-down split, a shared onset's digits splitting by pitch, chords staying one beat (with a stem, and with no stem found), same-direction stems folding into one chord, monophonic bars with flipping stems staying single-voice, whole-note placement, rest insertion front and back — plus a regression test pinning bar fullness on the reference file, and unit tests for reading a stem's direction from which end overhangs its noteheads.
- `server/tools/tab_extract/verify_tex.mjs` now skips the voices alphaTab pads unused bars with, so its beat count is what the transcription emitted; it also reports a voice count, and the round-trip test asserts more sounding voices than bars.
- Full suite in a clean container with the library mounted: **102 passed**.

### DIFF
--- a/server/fermata/glyph_rhythm.py
+++ b/server/fermata/glyph_rhythm.py
@@ -296,6 +296,12 @@ _SP = {
     # notehead <-> stem attachment
     "stem_x_tol": 0.68,           # was 3.5pt
     "stem_y_tol": 1.17,           # was 6.0pt
+    # How far outside a stem's own y-span an inner chord notehead may sit and
+    # still count as threaded onto it (see _stem_through_notehead). A third
+    # of a staff space is under half a notehead height, so it forgives the
+    # stroke ending a hair short of the outermost notehead's centre without
+    # reaching the next staff position.
+    "stem_span_slack": 0.35,
     # flags
     "flag_x_tol": 0.98,           # was 5.0pt
     "flag_y_tol": 1.76,           # was 9.0pt
@@ -934,9 +940,10 @@ DURATION_CODE = {4.0: 1, 2.0: 2, 1.0: 4, 0.5: 8, 0.25: 16, 0.125: 32}
 
 class NoteEvent:
     __slots__ = ("x", "y", "base_units", "flags", "dotted", "is_rest",
-                 "category", "notehead_kind", "tied_next")
+                 "category", "notehead_kind", "tied_next", "stem_dir", "stem_key")
 
-    def __init__(self, x, y, base_units, flags, dotted, is_rest, category, notehead_kind=None):
+    def __init__(self, x, y, base_units, flags, dotted, is_rest, category,
+                 notehead_kind=None, stem_key=None):
         self.x = x
         self.y = y
         self.base_units = base_units
@@ -946,6 +953,14 @@ class NoteEvent:
         self.category = category
         self.notehead_kind = notehead_kind
         self.tied_next = False  # best-effort: see _mark_ties()
+        # Voice signals. stem_key identifies the ONE engraved stem this
+        # notehead hangs off, so the several noteheads of a chord - which
+        # share a single stem - can be recognised as one beat rather than
+        # several. stem_dir is "up"/"down" (see _assign_stem_directions), the
+        # signal engravers use to separate an upper from a lower voice. Both
+        # are None for a notehead with no stem at all (a whole note).
+        self.stem_key = stem_key
+        self.stem_dir = None
 
     @property
     def quarter_units(self):
@@ -1009,6 +1024,38 @@ def _best_stem(stems, stem_xs, x0, x1, yc, tol, x_tol=None, y_tol=None):
     return best
 
 
+def _stem_through_notehead(stems, stem_xs, x0, x1, yc, tol):
+    """The stem this notehead is THREADED ONTO, for a notehead sitting part
+    way along a chord's shared stem rather than at its end.
+
+    _best_stem deliberately only attaches a notehead at a stem's END, which
+    is where the flag or beam that decides the duration lives, and keeping
+    that window tight is what stops a neighbouring voice's stem being read
+    for flags. But a chord is several noteheads on ONE stem, and every member
+    except the one at the stem's end sits far outside that end window - up to
+    an octave away. Those members still have to be recognised as the same
+    beat, so accept a stem whose x is beside this notehead and whose span
+    covers the notehead's centre.
+
+    This does not confuse the two voices of conventional polyphonic writing:
+    there the upper voice's stem goes UP and the lower voice's goes DOWN, so
+    each points AWAY from the other and neither spans the other's notehead.
+    """
+    lo, hi = _bounds(stem_xs, min(x0, x1) - tol.stem_x_tol, max(x0, x1) + tol.stem_x_tol)
+    slack = tol.stem_span_slack
+    best, best_dx = None, None
+    for i in range(lo, hi):
+        s = stems[i]
+        dx = min(abs(s.x - x0), abs(s.x - x1))
+        if dx > tol.stem_x_tol:
+            continue
+        if not (s.y0 - slack <= yc <= s.y1 + slack):
+            continue
+        if best_dx is None or dx < best_dx:
+            best, best_dx = s, dx
+    return best
+
+
 def _has_stem_near(stems, stem_xs, x0, x1, yc, tol):
     return _best_stem(stems, stem_xs, x0, x1, yc, tol,
                       x_tol=tol.rest_stem_x_tol, y_tol=tol.rest_stem_y_tol) is not None
@@ -1063,6 +1110,42 @@ def _beam_count_near(beams, stem, notehead_yc, tol):
         if y - clusters[-1] > tol.beam_level_gap:
             clusters.append(y)
     return len(clusters)
+
+
+def _stem_key(stem):
+    """Hashable identity for one engraved stem, so every notehead attached to
+    it can be recognised as belonging to the same beat."""
+    return (round(stem.x, 2), round(stem.y0, 2), round(stem.y1, 2))
+
+
+def _assign_stem_directions(notes, stems_by_key):
+    """Resolve each stem's direction ONCE, from every notehead hanging off it.
+
+    Direction cannot be read from a single notehead's own position on its
+    stem: a chord shares one stem that runs PAST all of its noteheads, so
+    the end further from any given notehead is on the wrong side for every
+    member but the outermost one - reading "the free end is below me,
+    therefore stem down" off the top note of an up-stemmed chord inverts it.
+
+    What identifies the direction is which end OVERHANGS the notehead group:
+    a stem sticks out roughly an octave beyond the note it points away from
+    and stops dead at the note it points toward. Page y grows DOWNWARD, so
+    the overhang above the group is (topmost notehead y - stem y0).
+    """
+    by_stem = collections.defaultdict(list)
+    for n in notes:
+        if n.stem_key is not None:
+            by_stem[n.stem_key].append(n)
+    for key, members in by_stem.items():
+        stem = stems_by_key.get(key)
+        if stem is None:
+            continue
+        ys = [m.y for m in members]
+        overhang_up = min(ys) - stem.y0
+        overhang_down = stem.y1 - max(ys)
+        direction = "up" if overhang_up > overhang_down else "down"
+        for m in members:
+            m.stem_dir = direction
 
 
 def _mark_ties(notes, curves, tol):
@@ -1188,8 +1271,10 @@ def decode_note_events(page, staff_top, staff_bottom, staff_x0, staff_x1, line_y
     dot_counts = _assign_dots(dot_owners, dot_events, tol)
 
     notes = []
+    stems_by_key = {}
     for ev in staff_events:
         if ev.category in NOTEHEAD_CATS:
+            stem = None
             if ev.category == "notehead_whole":
                 # whole notes never take a stem, flag or beam by definition -
                 # don't even look for one (a nearby unrelated stem in a dense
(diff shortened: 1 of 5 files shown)


sirged:
All five fixed in `e136ff1`. I reproduced each by driving `_build_measure_beats_glyph` directly before and after, so every fix is shown to bite. The theme you flagged was right for three of them, and it was also right about the fourth in a way I only found by measuring — see finding 5.

## 1. `_match_onset_columns` stole other onsets' columns

The extra-column search used the full notehead-to-digit window (`x_tol`, ~12pt), which is wider than the gap between consecutive columns. The first column is still found from the notehead x with `x_tol`; any *further* column must now sit within `_CHORD_SPLIT_SPACINGS` (0.6 notation spacings, ~3pt) of the column already claimed — enough for a chord's jittered split, well under the ~6.4pt that separates consecutive sixteenth columns.

```
before: [[(4,0,[(1,'7')]),(2,0,[])], [(4,0,[(5,'3')]),(2,0,[])]]   # (5,'3') on beat 1, x=110 note lost
after:  [[(4,0,[(1,'7')]),(2,0,[])], [(2,0,[]),(4,0,[(5,'3')])]]   # (5,'3') on beat 2
        unmatched_notes=1  # the digit-less notehead is reported, not hidden
```

## 2. `_stem_groups` merged consecutive onsets

Sharing a stem is necessary but not sufficient to be one beat — the noteheads also have to sound together. A stem's members are now split into runs within `onset_tol` of each other.

```
before: [[(4,0,[(2,'5'),(5,'3')]), (4,0,[(5,'7')])]]              _bar_quarters=2.0
after:  [[(4,0,[(2,'5')]), (4,0,[(5,'3')]), (4,0,[(5,'7')])]]     _bar_quarters=3.0
```

## 3. A rest could land in a voice already sounding

Took your second option. A rest matching no onset says nothing about which voice it belongs to or where it falls, so it is dropped and the silence recovered by `_pad_voice_to_budget` from the meter — the same reasoning the monophonic branch three lines above uses. Voices already holding a beat within `merge_tol` of the rest are excluded as well, which also stops two rests of one cluster stacking in one voice.

```
before: voice sums [5.0, 4.0] in 4/4   # melody [q,q,r,q,q], beats 3-4 a quarter late
after:  voice sums [4.0, 4.0]
```

## 4. A voice that lost every digit became pure inferred rest

Only voices that decoded at least one note are padded or emitted. A bar of nothing but decoded rests still keeps them, so genuinely silent bars are unaffected.

```
before: [[(4,0,[(1,'7')]),(2,0,[])], [(2,0,[]),(4,0,[])]]   inferred=5.0
after:  [[(4,0,[(1,'7')])]]                                 inferred=0.0
```

`inferred_rest_quarters` drops exactly where the phantom voices were: Katawaredoki 26q → 0q, and its multi-voice count 46 → 30. To Zanarkand's 2.5q is unchanged, so the genuine inferred rests are still reported.

## 5. The half-note stem lookup — measured, and the assumption was wrong

I tried the tight attachment window first and it was a bad fix, so I measured instead of guessing. Over 2004 half noteheads with a stem in the library, the distance from the notehead centre to its **own** stem's near end is median 0.44 and max 0.94 staff spaces. A 0.5-spacing window loses 46% of real attachments; library-wide it cost 17 exact bars and 30 quarters on the seven reference files. y-distance cannot separate own-stem from neighbour-stem, because the two populations overlap.

What does discriminate is the **side**: an up-stem leaves a notehead at its right edge, a down-stem at its left. Measured agreement is 99.68% on stemmed filled noteheads and 97.65% on half notes (21,700 of them). So `_assign_stem_directions` now drops a stem whose side contradicts the direction its overhang implies, rather than believing it — a wrong direction inverts the voice, no direction only loses it. Taken over the group's mean x, so the one notehead a second-interval chord displaces to the far side cannot throw the chord's stem away.

```
left-edge stem whose overhang says "up"  -> dir=None, key=None   (dropped)
its own right-edge up-stem               -> dir=up               (still believed)
genuine left-edge down-stem              -> dir=down             (still believed)
```

**This does not cover your exact scenario, and I am not claiming it does.** A neighbouring voice's down-stem sitting where this notehead's own down-stem *would* sit is geometrically indistinguishable from it, and still yields `dir=down`:

```
conventional-looking neighbour -> dir=down   (STILL WRONG)
```

So I took the fallback you offered for that part: the comment no longer asserts the opposite of what the code does. It now states what the lookup is for, the measured reason the window cannot be tightened, which subset the side check rejects, and that the consistent-looking case remains and shows up in the overfull-bar count.

## Metric sweep — this trades metric for correctness, and I would rather say so than pick the flattering number

Three of the five bugs were *inflating* the metric: column theft dropped notes from robbed onsets, onset merging removed a duration, and the phantom voices/rests padded voices to their budget. Fixing them gives back some exactness.

| library (297 files, 291 extractable, 0 crashes) | main | PR as reviewed | after these fixes |
|---|---|---|---|
| bars summing exactly | 1814 (17.0%) | 7916 (74.1%) | **7741 (72.4%)** |
| overfull bars | 8017 | 2177 | **2252** |
| total absolute error | 18707.06q | 3855.44q | **4123.81q** |
| bars emitted as two voices | 0 | 7793 | 7606 |

Isolating fix 5 (side check disabled, fixes 1–4 in place): 7761 exact (72.6%), 4097.31q. So of the 1.7-point drop, **0.2 points and 27 quarters are the side check** and the remaining 1.5 points / 242 quarters are the genuine bug fixes returning metric that was not earned.

Your confirmed baseline is intact — **To Zanarkand is unchanged at 46/50 exact (92.0%), 6.00q, 35 multi-voice bars.** Reference set:

| File | exact | error | multi-voice bars |
|---|---|---|---|
| To Zanarkand (Final Fantasy X) | 46/50 (92.0%) | 6.00q | 35 |
| Dream Fragments (Chrono Cross) | 28/35 (80.0%) | 15.50q | 25 |
| Katawaredoki (Your Name.) | 23/69 (33.3%) | 76.00q | 30 |
| Nightsong (Baldur's Gate 3) | 7/20 (35.0%) | 15.50q | 12 |
| 1 AM (Animal Crossing New Leaf) | 12/17 (70.6%) | 10.50q | 8 |
| Hinata vs Neji (Naruto) | 17/27 (63.0%) | 15.75q | 24 |
| Going Home (Bleach) | 21/39 (53.8%) | 13.50q | 0 |
| **total** | **154/257 (59.9%)** | **152.75q** | **134** |

(was 48/257 = 18.7% and 472.75q on main). The only file that moved noticeably versus the reviewed PR is 1 AM, 7.50q → 10.50q, which is the side check.

All 287 emitted transcriptions still parse with the installed renderer.

## Tests

Regression tests added for findings 1–4 by name — `test_an_onset_short_of_digits_does_not_eat_the_next_onsets_column`, `test_one_stem_shared_across_two_onsets_is_not_one_chord`, `test_a_rest_matching_no_onset_is_dropped_not_added_to_a_sounding_voice`, `test_a_voice_whose_notes_all_lost_their_digits_is_not_emitted`, plus `test_a_bar_of_nothing_but_rests_still_keeps_them` guarding the finding-4 filter — and for the side check, `test_a_stem_whose_side_contradicts_its_overhang_is_not_believed` and `test_a_seconds_chord_keeps_its_stem_despite_one_displaced_notehead`.

Two of my earlier stem-direction tests had put the stem at the notehead's centre-x, which real engraving never does; the side check caught that and they now use a real edge position.

Clean container, library at `/testlib`, node and alphaTab present so nothing skips:

```
platform linux -- Python 3.12.14, pytest-9.1.1, pluggy-1.6.0
collected 109 items
...
============================= 109 passed in 5.39s ==============================
```

PR #53, opened 2026-08-20
Emit MusicXML and document the tab profile

sirged:
Transcription emitted alphaTex, the renderer's own text format. Only this
project reads it, and its bar arithmetic could only be checked by a script
written here. MusicXML is the canonical output now, and the encoding is written
down as a profile — `docs/musicxml-tab-profile.md` — so another program can
produce and consume the same thing.

**It is a documented subset of MusicXML 4.0, not a new format.** No invented
elements, no container of our own, thirteen numbered conformance rules and two
examples that are validated by the test suite rather than typed out by hand.
MusicXML already specifies tablature — `<staff-details>` with `<staff-tuning>`
per string, `<technical><fret>` and `<string>` on the note — so the value is in
removing the ambiguity the specification deliberately leaves.

## The rules, in brief

| | |
| --- | --- |
| 1–3 | `score-partwise` 4.0, **no DOCTYPE**; `<divisions>` declared once and every duration an exact integer multiple of it; one part |
| 4–5 | Opening `<attributes>` in the schema's own order, with a TAB clef, `<staff-lines>` and one `<staff-tuning>` per string; **`line` counts staff lines from the bottom**, so `line = staff-lines + 1 - string`; `<capo>` after the tunings, tunings written un-capoed |
| 6–8 | Voices written one at a time, numbered from 1, each after the first preceded by a `<backup>` returning to the measure start; chord members carry `<chord/>` and do not advance the position; **every sounding voice's durations sum to the measure's duration** |
| 9–13 | `<fret>` and `<string>` on every sounding note, frets from 0 and strings from 1 for the highest; `<pitch>` present and consistent with string + fret + tuning + capo; pitch within what `<octave>` can express; spelling decided by the key signature on the line of fifths, ties to the plain letter then the flat |

Most of what can go wrong is element ordering — `<attributes>`, `<note>`,
`<staff-details>` and `<staff-tuning>` are all `xs:sequence`, so `<voice>`
before `<type>` and `<clef>` before `<staff-details>` are mandatory and read as
unnatural. Rule 5 is the one that fails *silently*: staff lines are numbered
from the bottom and strings from the top, and inverting them mirrors every note
onto the wrong string in a file that still validates and still loads. There is
a test named after it.

## Rule 8 is why this format

Every voice's durations must sum to its measure. **MusicXML does not require
that** — the document says so — and it is stated as a rule here because it is
the property that lets an independent tool find a defective transcription.

The overfull-bar count stops being a self-check. music21 reads 2948 defective
measures of 10689 across the library, which is *exactly* what the extractor
reports about itself; that agreement is now measured rather than assumed.
Under-full bars are counted too, which they were not before — a note dropped
for want of a fret number used to leave a bar looking clean.

Nothing is padded or trimmed to make the sums come out. A bar that does not add
up is emitted as it was read, because smoothing it over would hide the defect
the rule exists to expose.

## Pitch spelling

String, fret and tuning give an exact MIDI number; MusicXML wants `<step>`,
`<alter>` and `<octave>`, and only the key signature decides between F sharp
and G flat.

The key signature is now decoded from the accidentals the score engraves
**between the clef and its printed meter**. The meter is what makes the reading
safe: an accidental belonging to the first note sits to the *right* of it, so
without that boundary a piece in C major opening on an F sharp reads as G
major. Since engravers print the meter once and reprint the key on every
system, this answers on the first system and declines elsewhere, which is all a
document-level key needs. It resolves for 280 of 291 scores, and the values
look like guitar music — 97 with no accidentals, 91 with one sharp, then 34, 21
and 17 for two, three and four.

Spelling is settled on the line of fifths: of a pitch class's two
single-accidental spellings, take the one nearer the key's centre at
`fifths + 2`; on a tie prefer the spelling needing no accidental, then the
flat. Both halves of that tie-break earn their place — A flat rather than G
sharp in C major, and E natural rather than F flat in A flat major. Every note
of every key comes out spelled the way that key spells it (there is a test over
nine keys, and an exhaustive round-trip over all 128 pitches × 15 keys).

**What remains guessed**, stated in the document rather than annotated per
note: a chromatic note in a key of five or more accidentals can come out
unconventionally, B sharp rather than C natural in B major. It is the right
pitch, spelled oddly. And where no key can be read, `<fifths>0</fifths>`. Both
cost only how accidentals are written — the key never changes which pitch
sounds, which is why a documented default is the honest answer here instead of
a per-note uncertainty record.

## Three details worth flagging

- **No DOCTYPE.** The MusicXML DTDs are deprecated as of 4.0, the public DTD
  URL no longer resolves, and secure-by-default parsers refuse a document that
  references an external DTD at all. The line costs interoperability rather
  than buying it. Readers must still tolerate one, since plenty of real files
  carry it.
- **An impossible fret is not written as some other pitch.** A two-digit text
  span is occasionally two adjacent notes read as one, arriving as fret 78 — and
  78 semitones above a string is past the top of MusicXML's `<octave>` range.
  Writing it made five library files fail schema validation outright, which is
  far worse than one missing note. Such a note is omitted, its beat held open
  as a rest so the measure still satisfies Rule 8, and it is counted. **The XSD
  found this; no test written here would have.**
- **MusicXML replaces alphaTex as the stored format** rather than joining it —
  the unique index is `(score_id, source)` with no format in it, so two
  extracted rows could not coexist anyway. The `format` column is now read back
  and dispatched on instead of assumed, so rows written before this change keep
  rendering as the alphaTex they are and no data migration is needed. An edited
  row keeps the format it was edited in, and the guarantee that re-extraction
  never touches it is unchanged and still tested.

## Verification

Not self-consistency checks. Every claim below is from a tool this project did
not write.

| Check | Result |
| --- | --- |
| Real MusicXML 4.0 XSD (`lxml`) | **291/291 valid** |
| music21 parse | **291/291, no failures** |
| alphaTab (the renderer already shipped here) | **291/291 loaded** |
| Agreement between music21 and alphaTab | 10689 bars, 18295 voices, 96564 notes — identical |
| Rule 8, per music21 | 2948 of 10689 measures defective; extractor's own count agrees exactly, 0 mismatches over 291 files |

Library-wide: 297 PDFs, 291 converted, 0 errors, 6 not extractable (raster
scans and notation-only files, as before).

Tests: **176 passed, 0 skipped** in a clean container with the library mounted
and the real schema available — 56 of them new, covering child order for every
`xs:sequence` element, the string/line inversion, the Rule 8 arithmetic read
back out of the emitted XML, and pitch spelling. `pytest.importorskip` keeps
the schema step optional so CI stays hermetic; the ordering assertions are what
stand in for it there.

New: `server/tools/tab_extract/verify_musicxml.mjs`, alongside the existing
alphaTex one, which loads emitted files through the real alphaTab importer and
reports bars, voices, notes and the first note's MIDI value, string and fret.

## Deliberately out of scope

The document says this plainly rather than implying otherwise: bend contours,
tapping, hammer-ons, slides and harmonics, where implementation fidelity varies
too much to promise interoperability; tuplets, ties, grace notes, repeats and
beaming, which the extractor does not determine; printed `<accidental>`,
layout, lyrics and dynamics; multi-part and multi-staff scores; and mid-score
key changes. `<divisions>` is 480 partly so tuplets can be added later without
changing it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,16 @@ them.
 - **Tab out of a PDF** — engraved guitar PDFs carry their tab as real text and
   their rhythm in the music font's own glyphs, so Fermata reads both directly
   instead of guessing at pixels, and renders the result as a playable,
-  editable staff beside the original page. Fret numbers, strings and chords
-  come out reliably; how much of the rhythm survives depends on the engraving,
-  and whatever stayed uncertain is listed with the staff rather than hidden.
-  Music written in two voices is still flattened into one, so bars in
-  polyphonic pieces frequently do not add up yet — the count is reported.
-  Scanned PDFs have nothing to read and are not transcribed.
+  editable staff beside the original page. The transcription is written as
+  **MusicXML**, so it opens in other notation software rather than only here —
+  see [the tab profile](docs/musicxml-tab-profile.md) for exactly what is
+  written. Fret numbers, strings and chords come out reliably; how much of the
+  rhythm survives depends on the engraving, and whatever stayed uncertain is
+  listed with the staff rather than hidden. Bars whose voices the engraved
+  stems do not separate still do not add up — that is now a stated conformance
+  rule of the profile, so any MusicXML tool finds those bars too, and the count
+  is reported either way. Scanned PDFs have nothing to read and are not
+  transcribed.
 - **Practice tracking** — a session timer per piece, with recently-practised
   and neglected views so the library reflects what you are actually working on.
 - **Gig mode** — fullscreen, screen kept awake, oversized tap targets and
@@ -87,11 +91,24 @@ npm install
 npm run dev
 ```
 
+Tests:
+
+```bash
+cd server
+pip install -e ".[dev]"
+python -m pytest -q
+```
+
+Tests that need real sheet music read it from `FERMATA_TEST_LIBRARY` and skip
+when it is unset. Setting `FERMATA_MUSICXML_XSD` to a local copy of the
+MusicXML 4.0 schema additionally validates the emitted transcriptions against
+it — see [the tab profile](docs/musicxml-tab-profile.md#checking-a-file).
+
 ## Formats
 
 | Format | How it reads |
 | --- | --- |
-| PDF, engraved with a tab staff | Practice reader, plus transcription to a playable staff you can view beside the page |
+| PDF, engraved with a tab staff | Practice reader, plus transcription to MusicXML — a playable staff beside the page, and a file other notation software reads ([profile](docs/musicxml-tab-profile.md)) |
 | PDF, engraved without tab | Practice reader; nothing to transcribe from |
 | PDF, scanned | Practice reader only — a scan holds no text or glyphs to read |
 | MusicXML (`.musicxml`, `.mxl`) | Interactive: notation / tab / both, audio, full fidelity |
@@ -104,9 +121,9 @@ demo*) if your library is PDF-only so far.
 
 ## Roadmap
 
-- **Separating voices when transcribing** — the largest gap between a
-  transcription and a score you could practise from, and the reason bars in
-  polyphonic pieces overfill today
+- **Tuplets and ties when transcribing** — the largest remaining gap between a
+  transcription and a score you could practise from, and the reason bars still
+  overfill once voices have been separated
 - **Score creation and editing** — build a staff from scratch in the browser,
   instrument-agnostic with first-class guitar tablature support
 - **More import formats** — MIDI and plain-text tab in particular, since those
--- a/docs/examples/monophonic.musicxml
+++ b/docs/examples/monophonic.musicxml
@@ -1,0 +1,139 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<score-partwise version="4.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.musicxml.org/xsd/musicxml.xsd">
+  <work>
+    <work-title>Monophonic example</work-title>
+  </work>
+  <identification>
+    <encoding>
+      <software>Fermata</software>
+      <encoding-date>2026-08-19</encoding-date>
+    </encoding>
+  </identification>
+  <part-list>
+    <score-part id="P1">
+      <part-name>Guitar</part-name>
+      <score-instrument id="P1-I1">
+        <instrument-name>Guitar</instrument-name>
+      </score-instrument>
+      <midi-instrument id="P1-I1">
+        <midi-channel>1</midi-channel>
+        <midi-program>25</midi-program>
+      </midi-instrument>
+    </score-part>
+  </part-list>
+  <part id="P1">
+    <measure number="1">
+      <attributes>
+        <divisions>480</divisions>
+        <key>
+          <fifths>0</fifths>
+        </key>
+        <time>
+          <beats>4</beats>
+          <beat-type>4</beat-type>
+        </time>
+        <clef>
+          <sign>TAB</sign>
+          <line>5</line>
+        </clef>
+        <staff-details>
+          <staff-lines>6</staff-lines>
+          <staff-tuning line="1">
+            <tuning-step>E</tuning-step>
+            <tuning-octave>2</tuning-octave>
+          </staff-tuning>
+          <staff-tuning line="2">
+            <tuning-step>A</tuning-step>
+            <tuning-octave>2</tuning-octave>
+          </staff-tuning>
+          <staff-tuning line="3">
+            <tuning-step>D</tuning-step>
+            <tuning-octave>3</tuning-octave>
+          </staff-tuning>
+          <staff-tuning line="4">
+            <tuning-step>G</tuning-step>
+            <tuning-octave>3</tuning-octave>
+          </staff-tuning>
+          <staff-tuning line="5">
+            <tuning-step>B</tuning-step>
+            <tuning-octave>3</tuning-octave>
+          </staff-tuning>
+          <staff-tuning line="6">
+            <tuning-step>E</tuning-step>
+            <tuning-octave>4</tuning-octave>
+          </staff-tuning>
+        </staff-details>
+      </attributes>
+      <direction placement="above">
+        <direction-type>
+          <metronome>
+            <beat-unit>quarter</beat-unit>
+            <per-minute>80</per-minute>
+          </metronome>
+        </direction-type>
+        <sound tempo="80" />
+      </direction>
+      <note>
+        <pitch>
+          <step>E</step>
+          <octave>4</octave>
+        </pitch>
+        <duration>480</duration>
+        <voice>1</voice>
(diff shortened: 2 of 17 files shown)


sirged:
All six fixed in `127cb0b`. Both user-visible ones reproduced first, then shown fixed.

## 1 — MEDIUM: an edit stored under the wrong format never rendered

Confirmed. `saveEdit` sent `transcription?.format` — the format of the row that was *loaded* — and the server trusts an explicit `format` over its own sniff, so pasting alphaTex into the source editor of a MusicXML transcription stored it `format='musicxml'`, the viewer dispatched on that label, handed alphaTex to the MusicXML loader, and nothing appeared.

The client now sends **no** `format`; the server reads it off the content, which is the only thing that knows what was typed. Sniffing client-side as well would have been a second copy of the rule with its own chance to disagree. The endpoint still accepts an explicit format for a client that genuinely knows.

The sniff is also stronger: `content.lstrip().startswith("<")`. The old prefix list missed a document opening with a comment, and alphaTex has no form that can begin with `<` — metadata lines begin with a backslash, beats with a colon or a fret number. Covered by a parametrised test over eight inputs, plus a round-trip test that pastes alphaTex over a MusicXML row and MusicXML back over that.

## 2 — MEDIUM: key signature read as C major, high confidence, under an octave-transposing clef

Confirmed, and worse than latent — the library contains the trigger. `Classical/PrimoGuitar Misc/Sagreras-Leccion-84.pdf` carries an OpusSpecial `digit8` at x 60.5–66.7, inside the clef's own span of 57.1–73.0: an octave-transposing treble clef.

Reproduced against the old rule on a synthetic staff (transposing clef, 4 sharps, stacked 4/4):

```
staff: transposing treble clef (8 under the clef), 4 sharps, then a stacked 4/4
  OLD rule (any digit ends the key signature): (0, 'no accidentals between the clef and the meter')
  NEW rule (only a stacked pair ends it):      (4, 'meter_x0=110')

stray numeral between the 2nd and 3rd sharp:
  OLD: (2, '2 accidental glyph(s)')
  NEW: (4, 'meter_x0=110')
```

Zero, not `None` — a positive answer, reported as `glyph-decoded` with confidence "high — read from the accidentals engraved between the clef and the meter". E major silently spelled as C major. And four sharps read as two when any numeral lands mid-run.

Fixed as you suggested, by borrowing the guard `decode_time_signature` already had. A meter is now only a common/cut-time symbol, or a numerator digit run **stacked** over a denominator run — and both readers now share one function, `_stacked_digit_pairs`, so they cannot drift apart on what a printed meter is. `_signature_from_window` keeps its exact previous behaviour and diagnostics.

Two further guards:

- Where no such meter is found, the answer is **"not detected"**, not zero. "No sharps or flats" is an answer; "I could not look" is not, and they must not come back the same.
- The clef wall is the **leftmost** clef's right edge. `max()` over clef x-values meant a courtesy clef further along a short system could push the wall past the accidentals and empty the window.

Eleven tests over synthetic glyph windows, including the transposing clef, the stray numeral, a lone meter digit, a meter left of the clef, and the genuine empty key signature that must still read 0.

**Why this file still reads correctly today, and why that is not reassurance:** its transposing 8 sits 4.62pt below the staff bottom — **0.93 staff spaces**, just outside the 1.0-spacing search band. The correct reading survived by less than one staff space of engraving luck. A different staff size or engraver puts it inside.

## 3 — LOW-MEDIUM: short bars now downgrade confidence

Confirmed and fixed. Rule 8 treats both directions as defective and the confidence string now does too.

The threshold counts a bar **once** however it is wrong. A polyphonic bar can have one voice over its meter and another under it, so `overfull + short` double-counts it and can exceed the bar count — a real hazard for a naive `>= 0.25` test. `_bar_conformance` is now a `NamedTuple` with a `defective` field, and that is what the downgrade uses. Tested at the boundaries, including that 2 doubly-wrong bars in 10 is 20% and must *not* downgrade.

While there I exposed the four counts as data — `bars_overfull`, `bars_short`, `bars_defective`, `bars_measured` — on the extraction result and the API response, rather than only as prose in the warning list. That is what let the cross-check below compare numbers instead of scraping sentences, and it is documented.

## 4 — LOW: the document's accidental bound was off by one

You are right, and it is normative, so it mattered. Scanned every key from −7 to +7 for spellings that are exotic *and* non-diatonic in that key:

```
fifths -6: ['Fb']      fifths +4: ['E#']
fifths -5: ['Fb','Cb'] fifths +5: ['B#','E#']
fifths -4: ['Cb']      fifths +6: ['B#']
fifths -3 .. +3: none
```

Threshold is **four**, exactly as you said. Corrected in `spell_pitch`'s docstring and in the document, with both examples named and the octave shift called out (C flat 5 and B natural 4 are the same pitch). The library has 17 scores at four accidentals and 3 at five, so this describes real output.

## 5 — LOW: capo default disagreement

Fixed. `unrepresentable_notes(measures, tuning, fifths=0, capo=None)` now mirrors `build`'s parameters in the same order with the same defaults, and the docstring says why: both shift which notes are representable at the octave boundary, so a caller passing one and not the other would get a count that quietly disagrees with the file.

## 6 — LOW: reported note count overstated the file

Fixed. `notes` is now the written count. It was 96570 across the library against the 96564 the file actually holds; both sides now read **96564**, confirmed independently rather than by construction. Added a test tying extracted − unwritable to what the emitted XML contains, and a second one asserting the reported conformance counts match what a consumer computes from the document.

## Verification

Library-wide re-run — 297 PDFs, 291 converted, 0 errors:

| Check | Result |
| --- | --- |
| Real MusicXML 4.0 XSD (`lxml`) | **291/291 valid** |
| music21 parse | **291/291, no failures** |
| alphaTab | **291/291 loaded** |
| music21 vs alphaTab | 10689 bars, 18295 voices, 96564 notes — identical |
| Rule 8: extractor vs music21 | **2948 vs 2948 defective bars, 0 mismatched files** |
| `notes`: extractor vs music21 | **96564 vs 96564, 0 mismatched files** |

Key-signature distribution **unchanged** — 280 of 291 decoded, same spread (0:97, +1:91, +2:34, +3:21, +4:17, +5:3, −1:14, −2:6, −3:7, −4:1). Nothing moved because the one file carrying the trigger misses the band by 0.93 staff spaces, as above. The guard converts that coincidence into a design; it is not a behaviour change on this library, and no score newly reports "not detected".

Reference score unmoved:

```
bars=50 notes=357 beats=307 fifths=1 key_source=glyph-decoded
MusicXML: measures=50 pitched notes=357 multi-voice measures=35
pitch vs string+fret+tuning disagreements: 0
first note MIDI: 64
alphaTab: {"bars":50,"voices":85,"beats":307,"notes":357,"dottedBeats":33,
           "firstNoteMidi":64,"firstNoteString":6,"firstNoteFret":0,
           "tuning":[64,59,55,50,45,38]}
```

Test output, clean container, library at `/testlib`, real XSD mounted:

```
........................................................................ [ 36%]
........................................................................ [ 72%]
........................................................                 [100%]
200 passed in 3.78s
```

200 passed, 0 skipped — up from 176, with 24 new tests. Frontend build passes.

PR #64, opened 2026-08-20
A rendering layer for the score view

sirged:
`TabViewer.svelte` configured the notation renderer inline, so appearance,
responsive layout and the renderer's quirks were spread through a component
that is mostly a toolbar. This puts all of it behind one module.

## The seam

`web/src/lib/score-render.js` (468 lines) owns every interaction with the
renderer. Nothing else in `web/src` imports it or names one of its types — the
only remaining mention outside that file is a comment in `vite.config.js` about
where the plugin copies fonts.

```js
createScoreView(host, {
  scroller, source, profile, preset, theme, transport,
  onReady, onPlaying, onError, onPassComplete, onLayout,
})
// -> { layout, theme, profile, preset, lastRenderMs,
//      setProfile, setPreset, setTheme,
//      setSpeed, setLooping, setMetronome, setCountIn,
//      playPause, stop, destroy }
```

Plus `SCORE_PROFILES`, `LAYOUT_PRESETS`, `SCORE_THEMES`, `PHONE_MAX_WIDTH`,
`TABLET_MAX_WIDTH`, `layoutForWidth(width, preset)` and
`readScoreTheme(name, root)`.

The vocabulary is deliberately Fermata's rather than the library's, because
VexFlow was the credible runner-up and stays credible while the authoritative
model is MusicXML on the server. `layout.mode` is `"page"` or `"horizontal"`,
not an enum. A source is `{kind: "alphatex" | "musicxml" | "file"}`, not the
byte-loader call the library happens to want. The transport is `setLooping` and
`setMetronome`, not volume fields on a live api object. Event names, settings
trees, colour objects, font objects and prototype quirks all stop at the file.

## Responsive layout

Layout is chosen from the width of the **score container**, not the window: a
side-by-side PDF comparison hands the staff half a desktop, and half a desktop
should lay out like the tablet it is that wide. Breakpoints are named constants
(`PHONE_MAX_WIDTH` 620, `TABLET_MAX_WIDTH` 1100) and re-applied inside the
renderer's own `resize` event, so a viewport change costs one render, not two.

There are two presets, because the same width wants different answers depending
on where the screen is:

| | phone (≤620) | tablet (≤1100) | desktop |
| --- | --- | --- | --- |
| `desk` | page, 1 bar/row, scale 1.15 | page, 2 bars/row, scale 1.05 | page, auto, scale 1.00 |
| `stand` | horizontal, scale 1.30 | horizontal, scale 1.20 | page, 2 bars/row, scale 1.15 |

`desk` is a pointer and a whole window, where more bars on screen beats size.
`stand` — what gig mode asks for — is a tablet on a music stand read from a
metre away with both hands busy, so glyphs grow and the score becomes one
endless system scrolled sideways: no page breaks to lose your place at, and a
single scroll axis a thumb or a pedal can drive.

The layer reflects its choices onto the host as `data-score-*` attributes, so
what it decided is readable in devtools and assertable in a test rather than
only visible in a screenshot.

## Theming

The staff is drawn from Fermata's palette. Tokens live in `web/src/app.css`
(`--score-ink`, `--score-line`, `--score-accent`, …) and the layer reads them
from the live stylesheet, so the palette keeps one home. Two themes: `parchment`
and `slate`, a dark staff for practising in the dark — drawn dark rather than
inverted, which is what the PDF reader has to do with a fixed page.

The renderer's playback cursor and range selection ship with **no colour at
all** — it creates their elements with position only, so they were invisible
before this. They are styled from the same tokens now.

`docs/rendering.md` lists exactly which resources the renderer exposes and,
more usefully, what is out of configuration's reach: there is no separate
note-head, stem or beam colour (every glyph follows main or secondary by which
voice it belongs to), there is no background colour and no background at all
(the surface is transparent), and the deprecated per-element font accessors are
silently ignored when passed as settings.

## The branding override

The renderer draws "rendered by alphaTab" onto every score from a private
method on its layout base class, called unconditionally from four render paths,
with no setting to disable it. The layer walks up the prototype chain from the
live layout object to the prototype that **defines** the method and replaces it
there, once. In 1.8.4 that owner is `ScoreLayout` for all three concrete
layouts — `PageViewLayout` and `ParchmentLayout` through `VerticalLayoutBase`,
`HorizontalScreenLayout` directly — so patching the owner covers every layout
mode rather than only the active subclass. None of those classes is exported, so
a live instance is the only route to them.

If the method cannot be found it warns to the console and carries on. Verified
by pointing the constant at a name that does not exist: one warning, no
exception, the score still rendered, and the only consequence was the
annotation reappearing.

Two honest costs, both in the doc:

- **It reaches past the public API.** The method is undeclared in the type
  definitions and its owner is unexported. It survives a stock Vite production
  build (property names are not mangled) — checked against the built output, not
  assumed. A version that renames it will trip the warning.
- **It needs the renderer on the main thread.** With the render worker on, the
  layout object does not exist in our realm and there is nothing to patch. So
  `core.useWorkers` is off, behind a `RENDER_IN_WORKER` constant. Measured, it
  is the faster choice anyway: the worker cost 130–260 ms of startup and
  re-rendered identically.

This is a runtime patch rather than a vendored copy with one method removed
because alphaTab is MPL-2.0 — modifying its files would oblige us to publish
the modified file.

## Verification

Driven in real Chrome, not inferred from a passing build.

- Demo route and a real transcription (a 27 KB MusicXML transcription of an
  engraved PDF, in the side-by-side comparison view) both render.
- Playback plays and pauses, the cursor advances and is visible; loop,
  metronome and count-in toggle; the speed select applies; the tempo ladder
  opens its controls, forces loop on and starts at its configured percentage.
- All three profiles render and change staff height (notation 230 px, tab
  213 px, both 321 px on the same score).
- **The annotation is absent from the rendered output** — asserted on the SVG
  text, not on the patch having run — at all three widths, in page and
  horizontal layout, in gig mode, and in the minified production build.
- Three widths produce three layouts (1 / 2 / auto bars per row at scale
  1.15 / 1.05 / 1.00; 6 / 5 / 3 partials; 943 / 708 / 321 px tall). Resizing
  back and forth through all three, twice, threw nothing and lost nothing.
- Themed colours reach the staff: `#241D0F`, `#B3A284`, `#6F6045`, `#8A6A24`
  in parchment and `#ECE2CD`, `#4D4331`, `#94866A`, `#C9A45C` in slate are the
  only fills in the output, and the title renders in Fraunces, fret numbers in
  Figtree.
- Gig mode still works: HUD appears, toolbar and header hide, preset flips to
  `stand`, and at tablet width the score becomes one 2246 px-wide system in an
  860 px viewport that scrolls sideways without losing partials. Escape
  restores `desk`.
- No console errors and no page errors in any run.

Timings after the change, medians on a real score:

| | |
| --- | --- |
| First render, renderer work | 6.9–16 ms |
| First paint, navigation to staff on screen | 267–403 ms |
| Re-render, renderer work | 0.5–1.0 ms |
| Re-render, to pixels | 16–35 ms |
| Resize-driven re-render, renderer work | 0.1–0.3 ms |

## Not changed

The demo route has never passed `gigMode` down to the viewer, so pressing F
there hides the header without showing the HUD. That predates this branch and
is left alone rather than folded into a refactor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,11 @@ them.
   interactively and can be switched between standard notation, tablature, or
   both staves at once, with audio playback, adjustable speed, and a moving
   cursor. The built-in synthesizer also drives practice tools: drag-select a
-  passage on the score to loop it, plus metronome and count-in toggles.
+  passage on the score to loop it, plus metronome and count-in toggles. The
+  staff is themed to match the interface, lays itself out differently on a
+  phone, a tablet on a stand and a desktop, and can be drawn dark for
+  practising in the dark — see [how scores are rendered](docs/rendering.md) for
+  what the renderer does, what that layer adds, and why this renderer.
 - **Tab out of a PDF** — engraved guitar PDFs carry their tab as real text and
   their rhythm in the music font's own glyphs, so Fermata reads both directly
   instead of guessing at pixels, and renders the result as a playable,
@@ -90,6 +94,11 @@ cd web
 npm install
 npm run dev
 ```
+
+Everything to do with drawing a staff — appearance, responsive layout, and the
+renderer's quirks — lives in `web/src/lib/score-render.js`; components never
+touch the renderer directly. [docs/rendering.md](docs/rendering.md) explains
+that boundary and the decisions behind it.
 
 Tests:
 
--- a/docs/rendering.md
+++ b/docs/rendering.md
@@ -105,11 +105,40 @@ layout all flow through it. The view it returns exposes `setProfile`,
 ### Responsive layout
 
 The layer picks `layoutMode`, `barsPerRow` and `scale` from the width of the
-**score container** — not the window, because a side-by-side PDF comparison
+**scrolling stage** — not the window, because a side-by-side PDF comparison
 hands the staff half a desktop and half a desktop should lay out like the tablet
 it is that wide. Breakpoints are `PHONE_MAX_WIDTH` (620) and `TABLET_MAX_WIDTH`
-(1100), and the choice is re-applied from the renderer's own `resize` event, in
-the handler, so a viewport change costs one render rather than two.
+(1100).
+
+Two details of this are load-bearing and were both got wrong first time round.
+
+**The width comes from the stage, never from the host.** The host is also the
+renderer's own container, and in horizontal layout it grows to fit the score
+(see the quirks below) — so measuring it would feed the layout's output back
+into its input: a wide score reads as a wide screen, flips to the desktop tier,
+shrinks, reads as narrow, and oscillates. A `ResizeObserver` on the stage, whose
+width never depends on the layout chosen, is what drives the decision.
+
+**A tier change is a full render of the layer's own.** The renderer's answer to
+a width change is resize-*optimised* rather than a re-layout, and it is not
+enough:
+
+- it rebuilds the layout only when `layoutMode` itself changed;
+- otherwise it asks the layout to resize, and `HorizontalScreenLayout` declines
+  — `supportsResize` is `false` and `doResize()` is empty — so **nothing is
+  drawn at all**. Crossing 620px in the `stand` preset is horizontal to
+  horizontal with only a scale change, which is exactly this case;
+- and the vertical layouts regroup systems only when `barsPerRow` is auto.
+  `_resizeAndRenderScore` tests `getBarsPerSystem(0) > 0` against the *new*
+  value and, when set, keeps the grouping it already has and merely refits
+  widths — so phone (1 bar) to tablet (2 bars) would stay one stretched bar per
+  row.
+
+Relying on that path meant the layer could announce a new scale it had not
+drawn. It now renders the tier change itself, from a microtask so the renderer's
+new width is already assigned, and before the next paint so no intermediate
+state is visible. On a 360-bar score that render is 29–61 ms; on a normal one it
+is 1–4 ms.
 
 There are two presets, because the same width wants different answers depending
 on where the screen is:
@@ -145,10 +174,14 @@ Two themes: `parchment` (warm paper, dark ink) and `slate` (a dark staff for
 practising in the dark). The PDF reader has to invert a fixed page; a rendered
 staff can simply be drawn dark, which is what `slate` does.
 
-Values must stay in a form the renderer's colour parser accepts — hex, `rgb()`
-or `rgba()`. It answers `null` for anything else rather than raising, and a null
-colour draws as nothing at all, so the layer validates each token and falls back
-to a built-in default.
+Values must stay in a form the renderer's colour parser accepts: hex, or the
+**comma-separated** `rgb()`/`rgba()` form. It splits the function body on
+commas, so the modern space-separated syntax — `rgb(36 29 15)` — is not one of
+them; it parses to `null`, and a null colour draws as nothing at all. Some
+malformed input makes it throw instead. So the layer does not try to recognise
+the syntax itself: it hands each token to the parser, catches both answers, and
+falls back to a built-in default for anything the parser will not take. A
+mistyped token gives a readable staff, not an invisible one.
 
 **Exactly what the renderer exposes** — this is the whole surface:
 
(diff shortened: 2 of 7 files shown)


sirged:
All five fixed in `d291420`. Every finding reproduced first, and each one was
traced into `dist/alphaTab.core.mjs` to confirm the mechanism before changing
anything. Two of them turned out to be worse than reported.

## 1 — resize did not re-lay-out

Confirmed exactly as described:

- `ScoreRenderer.resizeRender()` falls through to `layout.supportsResize`, and
  `HorizontalScreenLayout` has `get supportsResize() { return false }` with
  `doResize() {}` — so a horizontal-to-horizontal crossing logged *"Current
  layout does not support dynamic resizing, nothing was done"* and drew nothing
  while `publish()` announced the new scale.
- `_resizeAndRenderScore` opens with
  `const barsPerRowActive = this.getBarsPerSystem(0) > 0` against the new
  `barsPerRow`, and when true keeps `this._systems` and only calls `_fitSystem`
  — only the `else` branch rebuilds the grouping.

A tier change is now the layer's own `updateSettings()` + `render()`, queued on
a microtask so the renderer's new width is already assigned (`triggerResize`
sets `_renderer.width` *after* the event) and so the intermediate state never
reaches a paint.

### The second half of this one

Fixing finding 2 made the host grow to fit the score, and the host is also the
renderer's container — so measuring it fed the layout's output back into its
input. Observed live: a wide score read as a wide screen → desktop tier → page
layout → host shrinks → reads as narrow → horizontal → grows → oscillating, with
`data-score-renders` climbing on its own. The width now comes from a
`ResizeObserver` on the scrolling stage, whose width never depends on the layout
chosen. That is also the more correct reading of "available space".

## 2 — the score was drawn off its paper

Two separate problems behind that one symptom.

The paper stopping short is fixed. Worth recording that the first attempt was a
`.at-host[data-score-layout="horizontal"]` rule and **Svelte pruned it from the
stylesheet** — the attribute is written from JS, so the compiler decided the
selector could not match. It is computed in the layer now, which also removed
the class/render ordering skew.

The bigger problem was underneath: the renderer sizes `.at-surface` from the
`totalWidth` it reports (2222 px) but draws a partial wider than that (2631 px)
and clips the difference with `overflow: hidden`. Measured: **53 of 316 glyphs
past the clip** — the last bar and the final barline, with no scroll position
that could reach them, because the clip also bounded the scroll range. So
horizontal layout was quietly truncating scores.

The layer now measures what was actually drawn, sizes the paper to it and stops
the surface clipping — and clears that width again *before* any render, so the
renderer never measures the previous score's width instead of the screen's.
Partials land after `postRenderFinished` (measuring there returned zero SVGs)
and more arrive on scroll, so the measurement follows a `MutationObserver` on
the host, `childList` only so our own style writes cannot retrigger it.

Screenshot at the far end, parchment: bar 9 and the final double barline now sit
on paper with the stage gutter after them. Before, the last visible bar was 8 and
the staff lines ran off the card onto `--bg`.

## 3 — colour validation

Confirmed against `Color.fromJson`: it does `json.substring(start+1, end)
.split(",")` and only handles 3 or 4 parts, so `rgb(36 29 15)` returns `null`;
and it throws `FormatError` on a missing paren or a non-string. The regex is gone
— tokens go through the parser, and both `null` and a throw count as unusable.

Checked end to end with `--score-ink: rgb(36 29 15)` set on `:root`: the drawn
fills stay `#241D0F, #B3A284, #6F6045, #8A6A24` with 398 glyphs, no `fill="undefined"`.
Also exercised `oklch()`, a named colour, `var(--nope)` and empty — all fall back;
`rgb(36, 29, 15)` and `rgba(36, 29, 15, 0.9)` are accepted. `app.css` and the doc
now say comma-separated rather than implying `rgb()` is safe in general.

## 4 — theme name

`SCORE_THEMES.includes(name)`, matching how profiles and presets were already
validated. Tested against the real module:
`constructor`, `toString`, `valueOf`, `__proto__`, `nope`, `undefined`, `null`,
`42` — all resolve to parchment with 5 parsed colours, nothing throws.

## 5 — inert cursor width

Dropped, with a comment saying why the width belongs to the renderer.

## Re-verification

The resize test now compares **drawn geometry** either side of every boundary —
layout mode, bars per row, scale, partial count, widest partial, total height —
across both presets, both directions, twice. It asserts the shape is *identical*
within a tier and *different* between tiers, which is what the previous
"nothing threw" check missed.

| run | phone | tablet | desktop |
| --- | --- | --- | --- |
| desk / demo | page 1 @1.15, 6 partials, 943 px | page 2 @1.05, 5 partials, 708 px | page auto @1.00, 3 partials, 321 px |
| desk / transcription | page 1 @1.15, 5 partials, 486 px | page 2 @1.05, 7 partials, 702 px | page auto @1.00, 4 partials, 297 px |
| stand / transcription | horizontal @1.30, 2848 px wide | horizontal @1.20, 2631 px wide | page 2 @1.15, 7 partials, 768 px |

Three distinct shapes per run, each stable across four visits, in the dev server
**and** the minified production build. The `stand` phone/tablet pair is the case
that previously drew nothing: 2848 px vs 2631 px proves the new scale was
actually drawn, not just announced.

Gig mode is driven through the real preset switch; only `requestFullscreen` is
stubbed, because a fullscreen window cannot be resized over CDP.

Everything else re-run and unchanged: annotation absent at every width, both
layout modes, gig mode and the minified build; the guard still warns once
without throwing when pointed at a name that does not exist, and the annotation
returns as its only consequence; both themes' colours reach the staff; playback,
loop, metronome, count-in, the tempo ladder, the profile switch and gig mode
enter/exit all intact. No console or page errors in any run.

## Cost, remeasured

A tier change is a full render now, so the old 0.1–0.3 ms figure was measuring a
no-op. Real numbers:

| | |
| --- | --- |
| Tier-change re-render, normal score | 0.9–4.1 ms |
| Profile switch, library work | 0.5–1.0 ms |
| Profile switch, to pixels | 16–35 ms |

And on a 360-bar score, since a full render on every tier change is the thing
worth checking:

| | |
| --- | --- |
| First render, library work | 69 ms |
| First paint, wall clock | 103 ms |
| Tier-change re-render | 29–61 ms |
| Longest main-thread task | **102 ms** |
| Tasks over 250 ms | **0** (3 long tasks total) |

That matches the 106 ms / none-over-250 baseline, for the same reason — the
renderer chunks into partials and only draws what is on screen. No regression.

PR #65, opened 2026-08-20
Three staff themes, remembered in settings

sirged:
## Summary

- Replace the two staff themes (`parchment`, `slate`) with three: **Parchment** (default, unchanged), **White on black** (`noir` — true black surface, near-white ink, for a dim room or a bright stage) and **Black on white** (`print` — the printed-page look, most legible under harsh light). Tokens for each live in `app.css` alongside the existing ones; `score-render.js` picks them up the same way it already did.
- Add a server-side settings store: a `settings` table (key/value, so a new preference needs no migration) with a reserved `owner` column defaulting to a single-instance sentinel, so real accounts later are a data migration rather than a schema redesign. `GET /api/settings` and `PUT /api/settings` on the existing router, with defaults for a fresh install and rejection of unknown keys/invalid values.
- Add `web/src/lib/settings.svelte.js`, a small shared store that loads settings once and exposes them reactively; `TabViewer.svelte` reads the staff theme from it instead of local per-component state, so the choice applies wherever a score renders without any component knowing it's server-backed.
- Add a dedicated Settings view (`#/settings`, linked from the library sidebar) and remove the `◐` toolbar toggle that used to cycle the old dark theme — a persistent preference belongs in its own place, not a crowded toolbar.
- Fix a one-line bug: `Viewer.svelte` never passed `gigMode`/`onToggleGig` to `TabViewer` on the demo route, so pressing `f` there hid the header with no HUD to get back to (Escape still worked).

## Test plan

- [x] `server/tests/test_settings_api.py` — defaults on empty, write-then-read-back, upsert (not duplicate rows), unknown key rejected (whole write rejected, not partially applied), invalid value rejected, owner defaults to the single-instance sentinel, writing several settings in one call.
- [x] Full server suite run in a clean container with the real library mounted read-only at `/testlib`: **204 passed, 4 skipped** (skips are the pre-existing MusicXML-XSD tests, gated on an env var this run didn't set).
- [x] `npm run build` — clean, no warnings (an earlier pass had CSS incorrectly pruned as "unused" by Svelte's static analysis since the theme attribute is set via JS `dataset`, not template markup; fixed by wrapping those rules in `:global()`).
- [x] Playwright against the real backend (real library, real FastAPI, no route stubs) — 35/35 checks: all three themes reach the rendered staff with the exact expected surface and ink colours on both a real score and the demo route; the choice survives a full page reload and is visible from a brand-new browser context (no shared storage) hitting the same server, simulating a second device; the demo route's gig HUD now appears on `f` (bug fix); playback, loop, metronome, count-in, tempo ladder, profile switching and gig mode all still work.

## Screenshots

Settings view, and the same real score under all three themes, attached below.

### DIFF
--- a/docs/rendering.md
+++ b/docs/rendering.md
@@ -167,12 +167,20 @@ palette has one home:
 
 ```
 --score-surface  --score-ink  --score-ink-soft  --score-line  --score-accent
---score-dark-surface  --score-dark-ink  --score-dark-ink-soft  ...
+--score-noir-surface  --score-noir-ink  --score-noir-ink-soft  ...
+--score-print-surface  --score-print-ink  --score-print-ink-soft  ...
 ```
 
-Two themes: `parchment` (warm paper, dark ink) and `slate` (a dark staff for
-practising in the dark). The PDF reader has to invert a fixed page; a rendered
-staff can simply be drawn dark, which is what `slate` does.
+Three themes: `parchment` (warm paper, dark ink - the default, matching the
+rest of the interface), `noir` (true black with near-white ink, for a dim room
+or a bright stage) and `print` (black ink on white, the printed-page look,
+most legible under harsh light). The PDF reader has to invert a fixed page; a
+rendered staff can simply be drawn in whichever of these it needs.
+
+The staff theme is a user setting stored on the server (`/api/settings`), not
+browser-local, so it follows a person between devices - see
+`web/src/lib/settings.svelte.js` and the settings view
+(`web/src/lib/Settings.svelte`).
 
 Values must stay in a form the renderer's colour parser accepts: hex, or the
 **comma-separated** `rgb()`/`rgba()` form. It splits the function body on
@@ -223,8 +231,9 @@ invisible in page layout and so easy to ship broken:
 - The renderer draws the whole score as one system, far wider than any sensible
   card. Left alone, the staff scrolls off the paper onto the page background —
   and with the parchment theme that is dark ink on a dark page, effectively
-  invisible. `slate` hides the problem, because its surface is close to the page
-  colour.
+  invisible. `noir` hides the problem too, since its surface is close enough
+  to the page's own dark background; `print`'s white surface would have made
+  it obvious immediately.
 - The renderer sizes its drawing surface from the total width it *reports*, but
   draws partials wider than that and clips the excess with `overflow: hidden`.
   On a real transcription that hid 53 of 316 glyphs — the last bar and the final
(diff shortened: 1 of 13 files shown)


sirged:
Addressed all seven review findings in 47e2935.

**1 (stale load overwrites a save)** - `settings.svelte.js` now tracks which keys a local write has touched (`written`); the initial GET is only allowed to apply a key it hasn't been touched by a write, however late it lands. Reproduced against the pre-fix build: a GET delayed in delivery (but dispatched, and answered by the server, before a write) reverted the write from "Black on white" back to "Parchment" on the old code; on the fixed code the write survives.

**2 (rollback could write an unchosen value)** - a failed write now rolls back to a fresh `GET /api/settings`, not to whatever was in the reactive object before the optimistic write (which can be the pre-load default if the initial GET hasn't landed yet).

**3 (picker looked clickable mid-save)** - the whole grid disables while any write is in flight, not just the card being written.

**4 (failed save was silent)** - `Settings.svelte` now shows the error message next to the picker when a write is rejected.

**5 (stored values not validated on read)** - `get_settings()` now falls back to the default for a stored value that isn't (or is no longer) one of `SETTINGS_CHOICES`, and drops a stored key that isn't a known setting. Covers exactly the case this PR creates: an install with `staff_theme='slate'` from before the rename now reads back as `parchment` instead of highlighting no card.

**6 (theme list existed in two places)** - cross-referenced in both `api.py` and `score-render.js`, plus a new test that parses `SCORE_THEMES` out of the frontend source and asserts it matches `SETTINGS_CHOICES["staff_theme"]`.

**7 (removing the toolbar toggle regressed the use case)** - `TabViewer.svelte` now has a small theme picker next to the profile buttons (not in the transport row), writing through the same `settings.svelte.js` store. Reproduced against the pre-fix build: navigating to `#/settings` and back to change the theme tears down and rebuilds the `AlphaTabApi` (soundfont re-fetched, playback stopped) because `App.svelte` routes with `{#if}`. The in-viewer picker calls `setTheme()` on the existing view instead, so playback keeps running and the renderer is never rebuilt.

**Verification**
- Server tests, clean container, real library at `/testlib`: **208 passed, 3 skipped** (skips are the pre-existing XSD-schema tests).
- Playwright against the real backend: the original 35 assertions still pass unchanged, plus 7 new ones built specifically to reproduce findings 1 and 7 against the pre-fix build (extracted from the prior commit and built separately) before confirming they're fixed on this one - all 7 pass.
- Manually confirmed findings 3 and 4 with a slow/failing PUT: the non-clicked card disables during the write, and a rejected write surfaces its error text and rolls the selection back.

Not addressed here, out of scope per the coordinator's note: #66 (notation-only MusicXML crashes the renderer under the Tab profile).

PR #78, opened 2026-08-21
Give the practice metronome its own tempo

### LINKED ISSUE #60: Metronome at its own tempo, or a proportion of the score
The metronome currently only switches on and off, following whatever tempo the score declares. Practice needs it independent of that.

Needed:

- **Set a tempo directly**, in beats per minute, regardless of what the score says. Sometimes the marking is wrong, sometimes it is aspirational, and sometimes the passage being worked on wants a different speed than the piece.
- **Set a proportion of the original** instead — practise at seventy per cent and let the metronome follow. This is how tempo work is actually described, and it keeps its meaning when the piece changes tempo internally.
- Keep it audible and countable at slow speeds, where practice mostly happens.
- Make the current value visible while playing, so it is clear what you are hearing.
- Sensible subdivision and an accent on the beat, since a bare click is hard to place in compound meters.

Playback speed already exists and the tempo ladder already steps it, so this is about separating two things that are currently one: how fast the music plays, and what the click is counting. They usually agree and sometimes should not — a click at full tempo over playback at seventy per cent is a normal way to practise.

sirged:
## Summary

Closes #60.

The metronome used to be baked into the same generated MIDI as the notes, so it always followed whatever tempo the score declared, scaled by playback speed exactly like every other event. Practice regularly wants those two things to disagree - a click at full tempo over a passage slowed down for a hard bar, or a fixed BPM the marking has nothing to do with - and the renderer's own metronome setting cannot express that: there is no way to make its click run at a different tempo than the notes beside it.

**The renderer cannot drive this on its own.** `api.metronomeVolume` only changes how loud the *existing*, speed-coupled click is; there is no public setting that decouples its tempo from `playbackSpeed`. So the click is a second, independent audio path - a plain Web Audio oscillator scheduled off its own clock (the standard lookahead pattern, not per-click `setTimeout`, which drifts) - that never touches alphaTab's synthesiser or its notion of tempo. alphaTab's own metronome is muted permanently so the two can never sound at once; its count-in is a separate, untouched feature.

Two modes:
- **Proportion** (default) - a percentage of the score's own tempo, read continuously from `playerPositionChanged`'s `originalTempo` (the tempo alphaTab is internally playing before `playbackSpeed` is applied), so a piece that changes tempo mid-stream is tracked as it moves rather than resolved once at the start.
- **Fixed BPM** - a number set directly, ignoring the score entirely.

Time signature (for subdivision/accent) comes from a flat tick index built from `score.masterBars` once per load. Compound meters (6/8, 9/8, 12/8) click the eighth-note subdivision, accented every third click (the actual pulse), rather than a bare click on the dotted-quarter beat.

## Design decisions

- **Per-session, not persisted.** Mode/proportion/BPM live in `TabViewer`'s own state, same as `speed`/`looping`/`countIn` - none of those persist server-side either, and the server's settings schema (`SETTINGS_CHOICES` in `server/fermata/api.py`) only whitelists `staff_theme` today. Extending it was out of scope for a frontend-only change.
- **No tempo declared → alphaTab's own fallback (120 BPM).** `Score.tempo` already defaults to 120 when nothing is marked; a proportion taken of "nothing marked" degrades to the same fallback rather than inventing a second one.
- **Gig mode shows the live value, not the controls.** A stand is not where you pick a mode or type a BPM - those live in the full toolbar, set up before stepping into gig mode - but "the current value visible while playing" still has to hold at a distance, so a small `♩ 96` readout comes along into the HUD.
- **The count-in is untouched.** It's a separate, pre-roll-only feature; decoupling it from the score's tempo was outside this issue's scope.

## What I could not verify

I have no way to listen to audio output. I verified the click is real (a genuine `AudioContext`, real scheduled oscillator nodes) and that its *timing* is correct by measuring actual wall-clock gaps between scheduled clicks against the expected values - see the browser suite - and reasoned about the envelope/frequency choices (950 Hz plain tick, 1500 Hz + louder accent, ~50ms percussive decay) from the Web Audio API's own semantics. I did not, and cannot, confirm by ear that it sounds right. Worth a manual pass before merging: `/#/demo`, enable the metronome, play, then drop the speed to well below 1× and confirm the click doesn't slow down with it; try Fixed BPM at something very slow (e.g. 30) for audibility at practice speeds; and load something in 6/8 or 9/8 to confirm the accent is audibly distinguishable from the subdivision ticks.

## Testing

- `web/tests/unit/metronome.spec.js` (23 tests) - the pure tempo/subdivision/timing arithmetic in the new `web/src/lib/metronome.js`, no browser needed.
- `web/tests/browser/metronome.spec.js` (6 tests) - against a real alphaTab render and a real Web Audio click (a hand-built 6/8 MusicXML fixture, stubbed at the `/api/scores/*` routes the same way `score-compare-warnings.spec.js` already does). Every timing assertion reads real wall-clock gaps between real scheduled clicks (`data-metronome-clicks` mutating on the actual host element) rather than a value the UI merely intended to produce - the failure mode this project has been bitten by before.
- Raised the browser suite's test-count floor from 36 to 78 (49 existing + 29 added).

**Mutations run and confirmed red** (each reverted after confirming failure):
- `metronome.js`: `base * ratio` → `base + ratio` (proportion math) - 5 tests failed.
- `metronome.js`: dropped the `n > 3` guard in the compound-meter check - turned out to be genuinely dead code (3/8 produces the same output either way), so I removed it rather than adding a test that couldn't distinguish the branches; see the comment in `metronomePattern`.
- `metronome.js`: swapped `clampBpm`'s `min`/`max` - 10 tests failed.
- `metronome.js`: `<=` → `<` in the time-signature tick lookup - 1 test failed.
- `metronome.js`: swapped `secondsPerClick`'s `4/unit` for `unit/4` - 2 tests failed.
- `score-render.js`: multiplied the scheduler's click interval by `api.playbackSpeed` (reintroducing the exact coupling this issue removes) - caught by the fixed-BPM decoupling test immediately; the proportion-mode decoupling test didn't catch it on the first pass (its tolerance was too loose), so I tightened the bound until it did.
- `score-render.js`: swapped `metronomePattern(numerator, denominator)`'s argument order - the 6/8 accent-pattern test failed.
- `score-render.js`: made `setEnabled` stop resyncing the scheduler (simulating the old "just mute" behavior) - the on/off test failed.

## Test plan

- [x] `npm run build`
- [x] `npm run test:browser` (78/78 passing)
- [ ] Manual listening pass (see "What I could not verify" above)

### DIFF
--- a/docs/rendering.md
+++ b/docs/rendering.md
@@ -140,18 +140,106 @@ tempo. Two modes:
   *at the playhead*, read from `playerPositionChanged`'s `originalTempo` -
   the tempo alphaTab is internally playing at before `playbackSpeed` is
   applied - so a piece that changes tempo mid-stream is tracked continuously
-  rather than resolved once when playback starts.
-- `"bpm"` clicks a number set directly and ignores the score.
+  rather than resolved once when playback starts. That quarter-note-based
+  proportion is then converted onto the CURRENT bar's own click unit
+  (`unit/4`: an eighth-note meter clicks twice as often as a quarter-note
+  one) before anything is displayed or scheduled.
+- `"bpm"` clicks the typed number directly, as the click rate itself, in
+  every meter - the meter's own beat unit plays no part in the rate at all,
+  only in which clicks get accented. That is deliberately not the same
+  convention as `"proportion"`: it is what a physical metronome's dial
+  means, and what the issue asks for - a tempo set directly, regardless of
+  what the score says. 120 in fixed-BPM mode clicks 120 times a minute in
+  6/8 exactly as it would in 4/4.
 
-Time signature (for subdivision and which click is accented) comes from a
-flat tick-ordered index built from `score.masterBars` once per load, looked up
-against `api.tickPosition` on every scheduled click - not from alphaTab's own
-metronome MIDI events, which exist (`MidiEventType.AlphaTabMetronome`) but are
-timed on the same speed-scaled timeline this feature exists to get away from.
+Either way, **the number reported to a caller (`onMetronomeTempo`, and so
+the on-screen readout) is the same number that gets scheduled: clicks per
+minute, full stop.** `effectiveClickRate` in `metronome.js` is the one
+function that produces it, and both `tick()` and `report()` call it the
+same way with the CURRENT bar's own unit, rounded once and used unrounded
+nowhere else. An earlier version reported the quarter-note figure instead -
+96 for a 6/8 piece marked "quarter = 96" - while the eighth-note clicks it
+actually scheduled sounded 192 times a minute; showing the number a
+listener would have to convert in their head, rather than the number they
+are actually hearing, is exactly the "displayed and sounded disagree"
+failure this project keeps having to re-learn not to ship. The clamp
+(`MAX_METRONOME_BPM`) is applied to this same, already-converted rate for
+the same reason: clamping the quarter-note figure first and converting
+second would let a compound or unusually fine meter (4/128, or even a
+plain 6/8 at a fast marking) push the actual clicked rate far past what the
+constant - or the click's own envelope length - could survive.
 
-The count-in (`setCountIn`) is untouched by any of this - a separate,
-pre-roll-only feature that still uses alphaTab's own click at the score's
-tempo, scaled by `playbackSpeed`, exactly as before.
+Time signature and bar position (for subdivision and which click is
+accented) come from `api.tickCache.masterBars` - alphaTab's own lookup from
+generated-MIDI tick to the `MasterBar` sounding there - looked up fresh on
+every scheduled click, never cached across one. Two things about this are
+load-bearing:
+
+- **It has to be `tickCache`, not a hand-built index.** `api.tickPosition`
+  lives on the generated MIDI's timeline, which expands repeats and skips
+  unplayed alternate endings - the notated bar order and the played tick
+  order are different timelines the moment a score has one repeat sign in
+  it. An earlier version of this summed `MasterBar.calculateDuration()`
+  itself to build that mapping; past the first repeat, every position it
+  answered for was for the wrong bar - the wrong meter, the wrong accent
+  grouping, silently.
+- **The accent's phase is derived from the playhead every time, not carried
+  as a counter.** A `clickIndex` that only resets when the scheduler starts
+  drifts out of alignment the moment the metronome is switched on mid-bar,
+  the transport seeks, a loop whose length is not a whole number of click
+  periods wraps, or the meter changes mid-score. Recomputing "which slot of
+  the CURRENT bar is this" from `api.tickPosition` on every click instead
+  means all four of those are the ordinary case, not a special one to
+  detect: whatever the playhead is doing, the next click just answers the
+  same question again. One consequence worth knowing: when the click's own
+  rate doesn't evenly divide the bar's real duration (any proportion other
+  than 100%, or a fixed BPM), the accent does not recur at a fixed spacing
+  measured in clicks - it recurs at a fixed spacing measured in the music's
+  own bars, landing on or just after each real downbeat regardless of the
+  click's own tempo. That is the point, not a rounding error.
+
+Compound grouping (accent every third click) applies to 6/8-style meters
+written in sixteenths too (9/16, 12/16, ...), not only eighths -
+`metronomePattern` checks the denominator against both. x/4 meters are left
+alone even when the numerator is divisible by three (6/4): unlike 6/8,
+which is unambiguously two dotted-quarter pulses, 6/4 is genuinely
+ambiguous between two dotted-half pulses and six plain quarters, with no
+default worth guessing.
+
+One imprecision is worth stating rather than quietly living with:
+`api.tickPosition` answers "where is the playhead right now", not "where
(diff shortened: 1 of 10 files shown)


sirged:
Pushed a follow-up commit addressing both review passes. Summary of what changed, and the honest numbers.

## Fixed

**Repeat signs desynced the click's meter (the high-severity one).** The tick→meter lookup summed notated `MasterBar.calculateDuration()`, which is a different timeline than the generated-MIDI timeline `api.tickPosition` actually reports on the moment a score has a repeat sign - past the first repeat, the click read the wrong bar's meter, hence the wrong subdivision and accent grouping. Now sourced from `api.tickCache.masterBars`, alphaTab's own lookup, which already accounts for repeats and skipped alternate endings. Added a real fixture (two bars of 4/4 repeated, then 6/8) and a regression test that plays through the repeat and asserts the meter stays 4/4 for the whole second pass.

**Accent phase was counter-based, not playhead-derived (the other high-severity one).** The *grouping* (metre) was already read from the playhead on every click, but the *phase* (which slot in the bar) came from a counter that only reset when the scheduler started - wrong the moment the metronome is enabled mid-bar, the transport seeks, an uneven loop wraps, or the meter changes mid-score. Phase is now derived from the playhead every click, same as the grouping. Added tests for mid-bar enable and for a loop whose length isn't a whole number of click periods (the "three-bar loop at seventy per cent" case) - the latter proves the phase sequence can't be the naive "+1 mod N" a persistent counter would produce.

**The count-in overlap was real, not something my comment's "never sound at once" claim covered.** alphaTab raises its Playing state before a count-in even starts, so the practice click was sounding underneath every count-in at a different tempo. It now stays silent until the covering second Playing event that marks the count-in's actual end (no intervening Paused - that's what tells the two apart), then anchors fresh. Added a test that asserts zero real oscillators start during the count-in window.

**The readout showed a quarter-note figure while the scheduler clicked the meter's own subdivision - up to 4x apart in a compound meter, and my own 6/8 test had enshrined the mismatch.** Both the display and the scheduler now come from one function (`effectiveClickRate`), rounded once and used unrounded nowhere else, so the number on screen is always the number actually sounding - clicks per minute, in every mode and every meter.

**Smaller, all real:** compound accenting extended to sixteenths (9/16, 12/16); x/4 deliberately left simple even for 6/4 (genuinely ambiguous, documented as a decision); the clamp now applies to the *actual click rate* after the meter conversion, not a pre-conversion tempo, so an extreme meter can't push the real rate past what the click's envelope survives; gain levels now have headroom against clipping; a stall now floors to "now" and drops missed clicks instead of firing a burst (confirmed both ways - a deliberately-blocked-main-thread test, and a mutation removing the floor); `AudioContext` construction failing (Safari/iOS context limits) is now caught so it silences the click, not the Play button; `clampBpm` guards its own input instead of trusting every caller.

## Test suite: rebuilt around the real audio boundary

The previous suite measured wall-clock gaps between `onClick` callbacks - schedule-time bookkeeping, not the audio boundary. It would have passed unchanged if the lookahead scheduler were replaced with `setTimeout`, and collapsed toward meaningless at high click rates. Every timing and accent assertion now wraps `OscillatorNode.prototype.start` directly and reads the real `when`/`frequency` handed to Web Audio. Confirmed by deleting the actual `osc.start(time)` call (leaving the call site and all bookkeeping intact) and watching 8 of 12 browser tests go red on real timeouts waiting for oscillators that never start.

**Test count: 97** (49 pre-existing, 36 unit, 12 browser - up from the 78 in the first pass, since the unit suite grew substantially covering the tick-to-meter map, which had no coverage before). Floor raised to 97 in `tests/minimum-tests.js`.

**Two holes in the floor itself, both fixed:**
- `test.skip` was counted as if it ran (`suite.allTests().length` doesn't distinguish). Now counted via `onTestEnd`, executed-and-not-skipped only, regardless of pass/fail/flaky - so a genuinely failing suite can't also get a spurious "tests have gone missing" stacked on top of its real failure.
- A `--reporter` flag on the command line replaces Playwright's entire configured reporter array, which would silently drop the guard reporter without it ever running at all - no counting strategy inside a reporter can catch that, because the reporter itself never loads. Added `web/scripts/run-browser-tests.mjs`, now what `npm run test:browser` actually invokes: it owns the real `playwright test` call, strips any `--reporter` a caller passes, always forces its own reporter set (list + the guard + `json`), and checks the JSON summary itself afterward - entirely outside Playwright's reporter system. Verified: a plain run passes at 97; `npm run test:browser -- --reporter=list` still enforces the floor and reports the drop; `test.skip` on one test now fails the run; a genuinely failing test does *not* also trigger the floor message; `PLAYWRIGHT_ALLOW_PARTIAL=1` still opts out deliberately.

## Left alone, checked and not touched

Verified `n > 3` really was dead code (already removed, unrelated to this round) and that both Svelte compiler traps are clean.

## What I could not verify

Still no way to listen. The measurements in this round (from your own review passes) gave me confidence the scheduler itself is sound - exact 250.000ms/500.000ms gaps, zero standard deviation, zero clicks scheduled in the past under normal operation - which is more assurance than I could have produced myself. What I still can't confirm by ear: whether the accent (1500 Hz, louder) actually reads as more prominent than the tick (950 Hz) on real hardware, and whether the click's timbre itself sounds like a metronome rather than merely like two different beeps. Worth a real listening pass before merging.

Issue #39, opened 2026-08-19
Import the formats free score archives distribute, verifiably


sirged:
Rather than shipping adapters that fetch from freely-licensed archives, support the formats those archives distribute well enough that a file downloaded by hand imports cleanly and is verified to do so. Same benefit to the user, without reaching out to volunteer-run infrastructure, without making public-domain determinations on the user's behalf, and without adapters that break whenever a site changes.

## What the archives actually distribute

Mutopia publishes PDF, MIDI and LilyPond source. ClassTab publishes PDF, MIDI and plain-text tab. CPDL publishes PDF, MusicXML and MIDI. OpenScore and similar corpora publish MusicXML and MuseScore files. IMSLP is mostly scanned PDF with occasional MusicXML and MIDI.

## Current state, measured

- **PDF** — supported, including tab extraction from vector engravings.
- **MusicXML, compressed MusicXML** — supported and full fidelity. Verified importing.
- **Guitar Pro** — supported.
- **MIDI** — **not supported.** Verified: the renderer reports no compatible importer for a standard MIDI file. This is the largest gap, since MIDI accompanies almost every download on the archives above.
- **Plain-text tab** — not supported.
- **MuseScore** — not supported.
- **LilyPond source** — not supported.

## Worth adding, in order

1. **MIDI.** The widest gap and the most common companion file. Note honestly that MIDI carries pitch and rhythm but no string or fret assignment, so tablature has to be inferred from pitch — the same position-choice problem that makes automatic tab generation unreliable. Import should give accurate notation and playback, and mark any generated tab positions as inferred rather than authoritative.
2. **Plain-text tab.** Fret numbers laid out on string lines, which is structurally the same data the PDF tab extractor already produces. No fonts, no glyph decoding, no geometry — a small self-contained parser, and directly relevant to a guitar-focused library.
3. **MuseScore files.** A container holding the score as XML; worth reading for the corpora that publish in it natively, either by converting or by extracting an embedded score where one exists.
4. **LilyPond source.** Full rendering needs LilyPond itself, which is a heavy dependency for a small container. At minimum, read the title, composer and instrument out of the header so an accompanying PDF is labelled correctly instead of being named after its filename.

## Verification is the point

[...]

PR #79, opened 2026-08-21
Engrave the extraction fixtures CI was missing

### LINKED ISSUE #76: The extraction tests never run in CI
Continuous integration reports **36 skipped** server tests. Run locally with `FERMATA_TEST_LIBRARY` pointed at a real library, the same suite skips **3**.

The difference is every test that needs an actual engraved PDF. Those fixtures are personal sheet music — correctly gitignored and never committed — so the tests that depend on them cannot run anywhere but a developer's own machine:

- `test_tabextract.py` — 16 tests
- `test_transcription_api.py` — 7 tests
- `test_glyph_rhythm.py` — 2 tests

That is the wrong set to have uncovered. Reading tablature out of an engraved PDF is the thing this project does that nothing else does, and the glyph-decoding that gets rhythm out of the engraving is the most intricate code in it. Both are protected only by whatever a developer remembers to run locally, with the library present, before pushing. CI stays green either way — so a change that breaks extraction merges clean, and nothing says otherwise.

This is the same shape as the guard test that could not see a mutation putting every frequency an octave out: a suite that reports success while checking nothing. Here it is worse, because the skip is silent and the number only looks alarming if someone happens to compare it against a local run.

## What would fix it

Commit engraved fixtures that can legally live in the repository, and point the existing tests at them.

- **Engrave them rather than sourcing them.** A free engraver that produces vector PDFs with both notation and tablature gives fixtures whose licence is not in question, whose content is known exactly, and which can be regenerated. Purchased or downloaded arrangements can never be committed, which is how the gap arose.
- **Cover the shapes the code actually branches on**, not just one happy path: notation with tablature, tablature alone, two voices in one bar, tuplets, ties, harmonics, a non-standard tuning, an octave-transposing clef, and a bar that genuinely does not add up. Several of those currently have exactly one real-world example in the library, and one of them — a bar wrong in both directions at once — has none.
- **Keep the real-library tests.** They should stay, still skipped when the library is absent, because they are the only thing that exercises material this project did not generate itself. The point is to stop them being the *only* coverage.
- **Make the skip loud.** A summary line reporting how many tests were skipped for want of a library, so a green run cannot be mistaken for a complete one.

## The larger prize

With committed fixtures, the round-trip check becomes possible: engrave a score from known MusicXML, ingest the PDF, and compare the result against the input. That measures extraction accuracy against ground truth instead of against itself — which matters because self-consistency metrics cannot see the bugs that inflate them. Three of the five voice-separation defects found earlier made the bar arithmetic look *better* while making the music worse.

Worth doing before extraction accuracy work continues, since it is the only way to tell whether a change to that code helped.

sirged:
## What this changes

Closes #76. CI reported **36 skipped** server tests where a local run with `FERMATA_TEST_LIBRARY` skipped **3** — the difference being every test that needed an engraved PDF. Those fixtures were personal sheet music, correctly gitignored, so 16 extractor tests, 7 transcription-API tests and 2 glyph-decoder tests ran nowhere but a developer's machine.

`server/tests/fixtures/engraved/` now holds eight small scores, **engraved rather than sourced**, each committed with the MusicXML it was engraved from and a script (`server/tools/tab_extract/engrave_fixtures.py`) that rebuilds both. Their licence is plain, their content exactly known, and the extractor's answer is comparable against **what was asked for** rather than against its own previous output.

| fixture | shape |
| --- | --- |
| `notation_and_tab` | notation over tab, two pages, D major, dotted note, beamed 16ths, whole-note chord; every bar adds up |
| `rests_and_flags` | every rest value and every flag hook, including 32nds |
| `tab_only` | tab with no notation staff — the honest spacing fallback |
| `two_voices` | melody stems-up over accompaniment stems-down in one bar |
| `tuplet_and_tie` | a triplet (undetected, so its bar reads overfull) and a tie across a barline |
| `drop_d` | non-standard tuning named in the score's text, plus a metronome mark |
| `defective_bars` | bars over their meter, bars under it, and a bar wrong both ways at once |
| `notation_only`, `raster_scan` | the two things extraction must refuse |

### The empirical finding that shaped this

Investigated before building anything, since the decoder is calibrated to Finale's Maestro and Sibelius's Opus and a free engraver uses neither.

A MuseScore PDF run through the **real extractor**, before any change:

```
extractable = True                      # frets and bars read fine
bars = 2, notes = 7
time_signature_source = 'not detected (assumed 4/4)'
key_signature_source  = 'not detected (assumed no key signature)'
rhythm_provenance     = {'spacing': 1}
confidence[rhythm]    = 'low - inferred from note spacing only'
warnings: rhythm source: no readable music-font note glyphs on the paired notation staff
```

So: staff detection reached, fret-digit extraction reached, barlines reached, staff pairing reached — and glyph rhythm decoding, the time-signature decode, the key-signature decode and the Maestro fingerprint guard **all unreached**. An engraved fixture committed as-is would have covered a fallback while looking like it covered the decoder. That is the failure this issue exists to end, so **option 1 was not viable on its own**.

**Direction chosen: option 2 — calibrate a third font key.** It turned out unusually cheap, because MuseScore's PDF hands over something better than either existing key. Its Leland subset has the `post` table dropped *and* a glyph order minted per file — no names like Opus, no stable GIDs like Maestro — but the PDF's own **ToUnicode CMap carries the SMuFL codepoint**, which pymupdf already surfaces. A SMuFL codepoint is a meaning fixed by a published specification rather than by one exporter's convention, so it needs **no outline fingerprint**, and one table serves every conforming font — MuseScore's Leland *and* the Bravura it falls back to for glyphs Leland omits, on the same page. That also widens what can be ingested from free archives, which is #39.

Every entry was derived by engraving a score that asks for each symbol explicitly and reading back which codepoint appeared — stronger than reading a rendered outline, because the input says which symbol was requested. A bar of unbeamed 32nds yields exactly four `flag32` codepoints; ten one-bar meters yield each time-signature digit once. Diamond noteheads are **deliberately left unmapped**: two codepoints turned up for a half and a quarter diamond and which is which was not established, so a harmonic degrades honestly rather than decoding at a duration nobody verified.

### A second defect the fixtures exposed

Staff detection assumed a staff line arrives as one primitive per system. That is the exporter's choice, not a property of staves: MuseScore draws a **separate abutting piece per measure** (measured gap: 0.0pt), so a system of six narrow bars presented six pieces, none a quarter of the page wide, and the whole system was invisible. Detection depended on a score happening to have wide bars — a tab staff could be found on one system of a page and not the next. Collinear pieces are now joined before the length test at a sub-point tolerance, and the full-bleed rules a page draws at its own edges are dropped (they produced a phantom one-line "staff group" in every anomaly report).

## How it was verified

**Ground truth, not self-consistency.** Every beat's decoded duration is compared against the MusicXML the page was engraved from. `notation_and_tab` bar 1 was written `quarter, eighth, eighth, half` and comes out `:4 2.4 :8 4.4 :8 0.3 :2 2.3` — and those frets are right too: an `E4` written under the octave-transposing clef sounds `E3`, which is the 4th string at the 2nd fret. `rests_and_flags` recovers `:4 r`, `:8 r`, `:16 r`, `:1 r`, `:2 r`, `:32 r` and eight `:32` notes with every bar summing to 4.0. `two_voices` separates all 8 bars into two concurrent voices of 4 quarters each (flattened they would read 8 quarters against a 4/4 meter), confirmed through the real alphaTab importer as more sounding voices than bars. The decoder reports **zero** unrecognised glyphs across the fixtures.

**Real library unaffected.** With `FERMATA_TEST_LIBRARY` set, all 20 library tests still pass — including the ones pinning exact staff counts, note counts, bar conformance and `rhythm_provenance` on the reference score. That is the check that the staff-line joining and the new rest categories did not disturb Maestro decoding.

| | passed | skipped | library-dependent skips |
| --- | --- | --- | --- |
| CI (no library), before | 271 | 36 | 33 |
| CI (no library), after | 333 | 22 | **20** |
| Local (library), before | 304 | 3 | 0 |
| Local (library), after | 353 | 2 | 0 |

The remaining 2 skips in both are the MusicXML XSD tests (`FERMATA_MUSICXML_XSD`).

**Every test added or repointed was proven able to fail.** 25 mutations run, each reverted after; the ones that matter:

| mutation | went red |
| --- | --- |
| SMuFL map emptied | 20 tests |
| noteheadBlack mapped to something harmless | 9 |
| half / whole notehead read as filled | 7 / 3 |
| 32nd flag counted as an 8th; 8th flag not counted | 2 / 2 |
| half rest read as whole; 32nd rest read as quarter | 2 / 3 |
| augmentation dot read as a notehead | 4 |
| time-signature digit 4 read as a 5; sharps read as naturals | 1 / 2 |
| octave-transposing clef not classified as a clef | 1 |
| SMuFL range widened to every codepoint | 1 |
| one mapped codepoint enough to call a font "music" | 1 |
| staff-line pieces no longer joined (pre-fix behaviour) | 17 |
| join tolerance wide enough to weld ledger lines | 5, incl. 3 real-library tests |
| page-wide rules kept as staff groups | 1 |
| ties never marked | 1 |
| stem direction ignored | 9 |
| short bars not counted | 50 |
| "Drop D" not recognised in page text; tempo never read | 3 / 2 |
| raster page reported as vector | 2 |
| no-tab-staff score extracted anyway | 2 |
| non-extractable PDF accepted for transcription | 1 |
| a fixture PDF deleted | 24 (**failures**, not skips) |
| generator writes different MusicXML than is committed | 1 |
| the library-skip summary says nothing | 1 |

Two mutations survived the first round and both were my fault, not the mutations':

- *"the octave clef is not a clef"* survived because my assertion read `0xE052 in {e.code for ...}` — every glyph carries its codepoint whether the decoder understood it or not, so the assertion held with the clef classified as furniture. Exactly the guard-blind-to-the-one-thing-that-mattered shape. It now asserts the `(code, category)` pair, and dies on that mutation.
- *"a missing fixture skips instead of failing"* survived because the test used `pytest.raises(AssertionError)`, which lets `Skipped` propagate — the test itself would have skipped and reported nothing. Rewritten with an explicit `except BaseException` that converts a skip into a failure.

**The skip is now loud.** The run prints, in yellow: `20 test(s) skipped for want of a sheet music library - this run did NOT exercise extraction against real engraved scores`. It counts only library skips, not optional-tool ones (`node not available` says nothing about extraction coverage), and it is itself unit-tested. `Server tests` also installs alphaTab (runtime deps only) so the emitted transcription is checked against the importer that has to render it, and runs with `-rs`.

## Anything left uncertain

Stated in `test_engraved_fixtures.py`'s module docstring and the fixtures' README as well as here, because a fixture that looks like coverage and is not is worse than none.

- **Maestro and Opus still have no CI coverage.** Those fonts ship with Finale and Sibelius and cannot be committed either, so the glyph-ID fingerprint (both `test_glyph_rhythm.py` tests) and the Maestro end-to-end tests still need the library. What runs in CI is the third calibration.
- **Scale.** The reference score is 50 bars of real two-voice fingerstyle writing; the two-voice fixture is 8 contrived bars. A regression that only appears in density will still only appear against the library.
- **A tie across a system break is not matched.** `tuplet_and_tie` engraves two ties and only one is matched — the other is drawn as two partial curves with its notes on different staves. Asserted as a limitation rather than papered over, so a fix shows up as that test failing.
- **A bar wrong in both directions at once is engraved but cannot be reported that way.** The library has no example; this fixture does. What happens is that the short voice is padded to the meter with inferred silence — which is what stops voices drifting — so the bar reports overfull and **not** short. The padding is announced in the warnings. `_bar_conformance` handles the shape and has a unit test for it, but this is why extraction has not so far produced one.
- **Diamond/harmonic noteheads are unmapped on purpose** (see above). A score using them counts them as unrecognised vocabulary and degrades honestly.
- **A very short final system is still missed**, because a staff under a quarter of the page wide cannot be told from a beam by length. `tab_only` is 12 bars rather than 8 so its last system is one MuseScore stretches; an 8-bar version silently lost 2 bars. Unchanged behaviour, now understood and documented.
- **Re-engraving moves coordinates.** The PDFs were produced by MuseScore 4.6.3, named in the fixtures' README. A different version will shift geometry and can change what the tests measure, so regenerate deliberately. A test asserts the committed MusicXML still matches its generator, so a fixture cannot quietly stop being regenerable.
- **pymupdf version sensitivity.** The fixtures are fixed bytes, but the assertions run through `get_drawings()` / `get_texttrace()`; a future pymupdf that reports geometry differently would surface here. That risk already existed for the library tests.


sirged:
Follow-ups found while doing this, filed rather than folded in:

- #80 — a score engraved in any tuning other than standard or Drop D is transcribed as standard tuning, stated as fact in the MusicXML, at "high" fret confidence with no warning. Reproduced by engraving a real DADGAD score with the generator added here. The most serious of the three: a guess presented as a reading.
- #81 — ties are decoded (`NoteEvent.tied_next`) and then discarded; nothing emits them, and a tie across a system break is not matched at all. The fixture and test added here pin both halves of the current state.
- #82 — `instruments.spec.js:315` failed on the first run of this PR and passed on a plain re-run of the same commit, with no code change between. Pre-existing flake in the instruments spec (shared server, state leaking between tests); unrelated to anything in this branch, but worth not re-running past.

sirged:
## Second round: the four blockers, fixed and measured

All four checks green on the rebased branch. **Library sweep over all 297 files, main vs branch: 0 regressions, 15 files gain, 22 report an honestly capped confidence.**

### D1 — repeat brackets welded into a phantom staff line

Reproduced exactly before touching anything: the same 6 files, the same numbers (Star Ocean 5→3 staves / 24→14 bars / 179→102 notes, etc.).

**The tolerance was never the lever, and I can now say so with my own numbers.** Measured across the library: 7084 gaps get joined, 1418 of them exactly 0.0000pt, the widest that must join is **0.8543pt**; the closest that must *not* join is **1.0673pt**. Real clearance **0.213pt**. My earlier claim of 7.7pt was measured on one fixture page and did not generalise — the docstring now carries the library-wide figures. And my M14 mutation (gap 40.0) proved nothing about 1.0, exactly as you said; the suite was insensitive to a 16× widening.

**The guard is on the result.** A run assembled from more than one piece must have `STAFF_LINE_SIBLINGS_REQUIRED = 4` other rows spanning the same extent; a run that arrived whole passes through untouched. That last clause is why it leaves Moongazing alone — I confirmed its tab staff lines are single 513.8pt Finale strokes (1 piece), while the furniture at y=212.7 is 3 pieces. A bracket has no sibling at its extent; every real staff line has four or five.

Side effect worth noting: the branch now discards **5 fewer** of main's own 7-line groups (7 → 2), which is where 4 of the 15 gains come from — One Small Step +2 staves, Awake, Cuddle, Bonds of Sea and Fire +1 each.

**Reporting fixed independently of the cause**, as instructed. A discarded group now names the line counts, and a staff-sized one says plainly that a system's bars and notes are *missing rather than wrong*, and caps the fret confidence. That reaches **22 library files** whose systems main was already dropping in silence at "high" — the 10- and 12-line double-stroked groups you're filing as D11 are now audible even though the cause isn't fixed here.

Anomaly totals, library-wide: main **204** → branch **168** (one-line phantoms 156 → 125, seven-line 7 → 2). So the change now removes more phantom groups than it adds, reversing the +44 you flagged.

### D2 — margin-trimmed PDFs

Length ceiling replaced by a page-boundary test, priority noted as latent. The misleading refusal reason is gone.

### D3 — a codepoint alone was believed

I rebuilt your attack as a **committed fixture** (`fake_music_font.pdf`, synthesised by the script) and confirmed the full chain: pre-fix branch → `rhythm_provenance {'glyphs': 1}`, i.e. Times-Roman letters read as noteheads; main → refuses; post-fix branch → refuses identically.

Three requirements now stand between a codepoint and belief, each aimed at a named vector:

1. **the PDF must embed a font program** — kills the base-14 attack;
2. **the mapping must not be the synthetic identity `U+E000 + gid`** — kills the fallback-CMap producer vector directly, rather than hoping its arithmetic misses;
3. **at least one recognised codepoint must be a notehead** — under that identity mapping ordinary ASCII lands only on clefs, never a notehead, and a notation staff without noteheads has nothing here to decode anyway.

Critically, **the gate disabled SMuFL on zero real pages**: every library file that decoded through it still does, several with more systems than before.

**The test that pinned the hole open is gone**, replaced by one that asserts each requirement separately — and each requirement has its own mutation.

### D4 / D9 — the honesty gate was a ratio

`harmonics_dense` is engraved with **two** harmonics in eight dense bars, not two per bar, so the ratio genuinely cannot see them: the test asserts `unknown_ratio < _UNKNOWN_RATIO_WARN` **while** the decode is degraded anyway. One system degraded, one clean, 0 defective bars, codepoint named in the warning.

Paired with your other direction: unknown codepoints are classified by SMuFL block, so **noteheads, flags and rests gate confidence while expression furniture is reported and does not**. On real material this moved two systems each way — Mertz-Minka and Easy-Christmas stop being falsely degraded over furniture; Chopin and Obregon start being honestly degraded over noteheads that were previously invisible.

### D7 and D10

- **D7:** once any font on a page qualifies, every font's SMuFL-range codepoints are decoded or counted. The Chopin page's single Bravura harmonic is now counted — and it is a notehead, so that system degrades. That is one of the two new honest degrades above.
- **D10:** every fixture PDF's `/Creator` is asserted (`MuseScore Studio Version: 4.6.3`, or `fermata engrave_fixtures.py` for the two synthesised ones). I took your correction on the MusicXML guard and left it alone.

### Also done

- `SMUFL_RANGE` extended to **F8FF** — the optional block is no longer discarded unread.
- The skip count now comes from a **counter the fixtures increment**, not from matching words in a skip reason.
- **The short final system is pinned**, not prose: `tab_only_short_last_system` writes 8 bars, 6 are read, and the test asserts both that and the absence of any warning about it. A mutation lowering the length floor turns it red.

### Mutations

23 more, each reverted. Every test added in this round goes red under at least one. Four survived the first pass and all four were my fault, not the tests':

- two were no-op mutations (`None or (...)`, and a duplicated length check the single-anchor replace missed) — I refactored the duplication away so the floor is testable, then re-ran;
- two showed my weld test could not feel `STAFF_LINE_SIBLINGS_REQUIRED` or `STAFF_LINE_SIBLING_TOLERANCE`, so I added a second hostile page — one furniture row with exactly one companion at its extent, one nearly as wide as the staff — and both constants now have a test that dies when they move.

### Not touched, per your instruction

No new tuning coverage (#80 owns it, including the "Also available in Drop D" false positive). D5, D8 and D11 left alone; `defective_bars`' narrative now says the padding is what happens rather than presenting it as handled.

Issue #56, opened 2026-08-20
Move, rename, delete and reorganise scores


sirged:
The library is currently read-only: the scanner reads whatever folder structure exists and nothing can be changed from the application. Managing a collection means leaving the tool and moving files by hand, after which the scan has to catch up.

Needed:

- **Move, rename and delete** scores from the interface, with the change applied to the file on disk and the database record following it rather than being orphaned.
- **Reorganise** — put a score in a different collection, correct a composer, restructure a folder — without breaking the link to its transcription, practice history, tags or annotations.
- **Normalise on ingest**, optionally: bring files in under a consistent naming and folder scheme derived from their metadata, so a library assembled from several sources does not stay a mess.
- **Keep the source file and its generated MusicXML associated**, visibly, so it is clear which scores have a transcription and where each came from.

## This is the one destructive feature

Everything else in this project reads. This writes to a person's own files, and getting it wrong loses their library. So:

- **Dry run first.** Show exactly what will be renamed or moved, before anything happens, and make that the default for any bulk operation.
- Never delete a file as a side effect of an organisational change.
- Refuse to write outside the configured library directory, and test that.
- Make normalisation opt-in. A library someone has already organised the way they like must be left exactly as it is.

Prior art worth reading rather than reinventing: beets does exactly this for audio, MIT licensed, and its templated paths, dry-run posture and duplicate handling are directly applicable.

PR #99, opened 2026-08-22
Keep a library that reads as empty from erasing the record of practising it

### LINKED ISSUE #95: A library that looks empty deletes every score, and with it all practice history
If the library directory is empty when a scan runs, the scan deletes **every** score row — and three tables cascade from `scores`, so it also destroys every practice session, every tag, and every hand-edited transcription. Silently, automatically, with no confirmation and no undo.

## The chain, all verified on `main`

**1. A missing mount becomes an empty directory.** `ensure_dirs()` does `LIBRARY_DIR.mkdir(parents=True, exist_ok=True)`. So if the bind mount is not present — the host path was renamed, an external drive did not come back, Docker started before the mount was ready — the application does not fail. It creates an empty directory and carries on as though the library were genuinely empty.

**2. A scan runs by itself at startup.** The lifespan hook calls `init_db()` and then `scanner.start_scan()`. No user action is required, and nothing asks first.

**3. The scan deletes every row it did not see, with no floor:**

```python
existing = conn.execute("SELECT id, path FROM scores").fetchall()
for row in existing:
    if row["path"] not in seen_paths:
        conn.execute("DELETE FROM scores WHERE id = ?", (row["id"],))
```

With no files on disk, `seen_paths` is empty, so the condition is true for every row.

**4. Three tables cascade from `scores`** — `ON DELETE CASCADE` on practice sessions, tags, and transcriptions.

So: a container restart with the library unavailable wipes the user's entire practice record, their tags, and any transcription they corrected by hand. The scores themselves come back on the next good scan, because they are files. **Nothing else does.**

## Why this outranks everything else open

The deployment guide is explicit that the data in the config directory cannot be recovered from the PDFs, and tells people to back it up. That advice is sound and it is the only thing standing between this bug and total loss — but it only helps someone who took a backup *and* noticed before taking another one over the top.

Practice history is the point of the practice features. A person could lose a year of it because a drive did not mount.

## What a fix looks like

- **Refuse to delete when the evidence is implausible.** If a scan sees zero files while the database holds scores, that is not an empty library, it is a broken mount. Stop, change nothing, and say so. The same applies to losing a large proportion in one pass — a library does not usually shed most of itself between two startups.
- **Do not create the library directory.** Its absence is a configuration error and should be reported as one. Creating it converts a loud, obvious, harmless failure into a silent destructive one. The config directory is different — that one is ours to create.
- **Separate "the file is gone" from "the record is gone."** A score whose file has disappeared could be marked missing rather than deleted, which is also what makes a re-mount recoverable and would pair well with the content-hash relinking that already exists for renames.
- **Reconsider the cascades.** Practice sessions and hand-corrected transcriptions are original work that happens to be *about* a file. Tags likewise. None of them stop being true when the file moves. (#94 is already changing the sessions cascade for the same reason.)
- **Make destructive scan outcomes visible.** The scan reports a `removed` count into its state, but nothing surfaces "this scan removed 297 scores" as the alarming event it is.

## Reproduction

Point `FERMATA_LIBRARY` at a directory, scan it, record a practice session, then point it at an empty directory and restart. The scores return on the next correct scan; the session does not.

Filed against Rev 1 and ahead of the other Rev 1 bugs. Every other defect in this milestone produces a wrong answer that can be corrected. This one destroys data that cannot be reconstructed, on a trigger the user does not control and is not warned about.

sirged:
## What this changes

Closes #95. A library folder that read as empty made the startup scan delete every score row, and three tables cascaded from `scores` — practice history, tags, and hand-corrected transcriptions. This stops the scan from ever deleting a score row, refuses walks of the library it cannot believe, and stops creating the library folder.

- **`scores.missing_since`** — an absence is recorded as an absence, not as a deletion. The row and everything hanging off it stay put, which is what makes a remount recover by itself and is the only version of this that survives a trigger nobody has thought of yet.
- **A scan refuses implausible evidence.** Zero files while the database holds scores is refused categorically. Losing half or more of the still-present rows in one pass is refused above a floor of ten rows. Both report why, in `/api/scan/status` and in the log.
- **`ensure_dirs()` no longer creates the library folder.** Its absence is a configuration error and now reads as one. Under `restart: unless-stopped` that heals itself the moment the mount appears. The config folder is ours, and is still created.
- **A destructive scan says so** — `marked 5 of 12 score(s) missing`, at warning level, with paths.
- **The tags and transcriptions cascades were reconsidered and kept** — reasoning in `db.py`, and pinned by a test.
- Scan state: `removed` (always zero now, since nothing is removed) becomes `missing`, plus `restored`, `refused` and `refused_reason`. Nothing consumed `removed`.

## How it was verified

**The loss was reproduced first, on unmodified `main`, in a separate worktree**, with a real library, three practice sessions, a goal, tags and an edited transcription, driven through two separate processes so each stage reads its config fresh the way a container restart does. Counts are `score_tags` / `transcriptions(edited)` / sessions-naming-a-piece:

| Trigger | On `main` | With this change |
|---|---|---|
| Library becomes an empty folder | 2→0 / 2→0 / 2→0 — **total loss** | refused, **every count unchanged** |
| Library folder simply absent | scan proceeds against the folder it created | refuses to start, explains why |
| One file **edited and moved** in the same pass | 2→1 / 2→1 / 2→1 | 2→2 / 2→2 / 2→2, old row marked missing |
| Two identical copies whose shared folder is renamed | 3→1 / 3→1 / 3→1 | 3→3 / 3→3 / 3→3, both marked missing |
| Remount after a refusal | — | fully restored, `missing_since` cleared |

**One correction to the issue's second narrow trigger.** "Duplicate content where one copy is renamed" does **not** reproduce: the still-present sibling is filtered out by path, leaving exactly one candidate, and the relink works correctly. It takes *both* copies moving in the same pass to produce two candidates — which is what renaming the folder they share does, and that is the case in the table above. Worth knowing, because it is more ordinary than the version in the issue, not less.

**Also verified in a real container**, built from this branch: score scanned, practice logged, tag applied, transcription saved, container stopped, the file removed, container started. The scan refused, the log carried the reason, and the tag, the 2400 seconds and the transcription were all still there. Running the image with no library mount prints the configuration error and exits rather than starting.

**Tests.** 33 new tests in `server/tests/test_scanner.py`; suite goes 528 → 561 passing, with `FERMATA_TEST_LIBRARY` set so nothing silently skipped (the run prints `real-library tests all ran`). Full browser suite 146 passing, frontend build clean, container image builds.

**Every new test was proven able to fail: 25 mutations applied to the source, 25 killed, 0 survived.** Each test is named by at least one mutation. Among them: restoring the original `DELETE FROM scores`; removing the zero-files guard; `LOSS_FRACTION` to 0.9 and to 0.1; `LOSS_FLOOR` to 1; `>=` loosened to `>`; computing the refusal but not returning; clearing `missing_since` *after* the unchanged-file shortcut instead of before; dropping `missing_since=NULL` from the relink; silencing the warning, and making it always fire; restoring the `mkdir` in `ensure_dirs`; removing `missing_since` from `COLUMN_ADDITIONS`; dropping the `score_tags` cascade; counting already-missing rows in the denominator.

**The upgrade path** is tested from a database built by *the previous release's own code* (real `init_db`, with `missing_since` removed from `COLUMN_ADDITIONS`) holding real scores, sessions, a goal, a tag link and an edited transcription. Every row survives with its id, and the resulting `scores` table is compared column-for-column against a fresh install's — the #94 lesson about upgraded and fresh installs diverging.

**The `NOT IN` trap was checked, not assumed.** The only `NOT IN` in the codebase is `db.py`'s `score_id NOT IN (SELECT id FROM scores)`, which is already guarded by `score_id IS NOT NULL` and whose subquery selects a primary key, so no NULL can enter it. `missing_since` is a new column nothing reads, but a test drives the neglected view, the collections counts and the tag counts with a missing score present anyway, because that was exactly the shape of the last one.

## Anything left uncertain

- **There is no way to remove a missing score row.** Deleting the file used to do it; now it marks. There is no `DELETE /scores` endpoint, so somebody who really did delete a score is left with an inert row listed in their library until score management (#56) arrives. This is a deliberate trade — the alternative is a "forget the missing ones" action, which is the loaded gun this issue is about — but it is a real gap and #56 should close it, with the cascade consequences stated in the interface.
- **A library under the ten-row floor can be drained entirely, a little at a time**, without the proportional guard ever firing; only the zero-files test stands at the end of that road. Accepted because every step is a mark and one good scan restores it all, and written down in `scanner.py` rather than left to be discovered.
- **Missing scores are listed, not hidden.** "Your library is intact, these files are not reachable right now" is the true statement, and hiding them would show somebody whose drive has not come back the same empty library the bug produced. But nothing in the interface flags them yet — `missing_since` is in the score payload and unused. Worth a follow-up; deliberately not done here, since another change is in flight in `web/**`.
- **Two interface files were changed**, and only because this change invalidated them: `playwright.config.js` created its scratch library by relying on the server to `mkdir` it, and `viewer-practice.spec.js` cleaned up by deleting a file and waiting for the row to vanish. Neither touches the metronome work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,11 @@ home network.
 | `/data/library` | Your sheet music files (read + uploads)        |
 | `/data/config`  | Database, thumbnails, cache — back this up     |
 
+Fermata creates `/data/config` if it needs to, but never `/data/library` — a
+library folder that isn't there is usually a volume that didn't mount, and an
+empty one is indistinguishable from a library with nothing in it. It stops with
+an explanation instead, and recovers by itself once the mount appears.
+
 ## Development
 
 Backend (FastAPI, Python 3.12):
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -33,7 +33,9 @@ mkdir -p library config
 
 - `library/` is where your sheet music goes — PDFs, MusicXML, Guitar Pro
   files. Fermata only reads from and writes into this folder; it never touches
-  anything outside it.
+  anything outside it. It also will not create it: if this folder is missing,
+  Fermata stops and says so, because a missing library folder is far more often
+  a drive that did not mount than a folder nobody made yet.
 - `config/` is where Fermata keeps its own state: the database (titles, tags,
   practice history, transcriptions) and a thumbnail cache. You will not need
   to open this folder yourself, but you will back it up — see
@@ -155,7 +157,10 @@ docker compose up -d
 
 Your library files in `library/` are untouched by any of this — restoring
 `config/` brings back the database as it was at backup time, and the next
-startup scan reconciles it against whatever is currently in `library/`.
+startup scan reconciles it against whatever is currently in `library/`. That
+reconciliation cannot cost you anything: a score whose file is not found is
+marked as missing rather than removed, so its practice history, tags and
+transcription stay attached and come back with the file.
 
 ## Upgrading
 
@@ -306,10 +311,39 @@ Then restart:
 docker compose up -d
 ```
 
-There are two other reasons Fermata will stop on purpose, and because
-`docker-compose.yml` sets `restart: unless-stopped`, either one shows up as a
-container restarting over and over. Both print a plain explanation, so it is
+There are three other reasons Fermata will stop on purpose, and because
+`docker-compose.yml` sets `restart: unless-stopped`, any one of them shows up as
+a container restarting over and over. All print a plain explanation, so it is
 worth reading to the end of the log rather than only the last line.
+
+**"its library folder … is not there"** — the `library/` folder Fermata was
+told to use does not exist. This one is worth understanding rather than just
+fixing, because Fermata refuses it deliberately and used to do something much
+worse. A missing library folder is almost always a mount that did not appear:
+the host folder was renamed or moved, an external drive did not come back, or
+the container started before its volume was ready. Fermata will not create the
+folder for you — an empty library folder looks exactly like a library with
+nothing in it, and Fermata would have no way to tell the difference.
+
+Check that the folder is there, next to your `docker-compose.yml`, and that
+`docker-compose.yml` still has `./library:/data/library` under `volumes:`:
+
+```bash
+ls -la library
+```
+
+If the folder genuinely should exist and be empty — a first run before you have
+copied any music in — create it and start again:
+
+```bash
+mkdir -p library
+docker compose up -d
+```
+
+A restart loop here is harmless and self-healing: nothing has been changed, and
+the moment the mount appears the next restart attempt succeeds by itself. This
+is on purpose. Your practice history, tags and transcriptions in `config/` are
+not touched by any of it.
 
 **"this database is at schema version N"** — the database was written by a
 newer Fermata than the one you are running, so this one will not touch it. This
@@ -347,6 +381,27 @@ If that comes back empty, the files aren't where the container is looking —
 double check they're inside the `library/` folder next to your
 `docker-compose.yml`, not a `library/` folder somewhere else, and that
 `docker-compose.yml` still has `./library:/data/library` under `volumes:`.
+
+If the listing comes back empty **and** Fermata was already showing your
+scores before, look in the log instead:
+
+```bash
+docker compose logs fermata --tail 50
+```
+
+You should find a line saying the scan "did not reconcile the library", with a
+reason. Fermata refuses to act on a walk of the library it cannot believe — a
+folder that suddenly holds no readable score files at all, or one that has lost
+more than half of what it had, is treated as a mount problem rather than as news
+about your music. Nothing is changed when that happens, and the message says so.
+Fix the mount and scan again; everything comes back.
+
+If some of your scores are listed but marked as missing, that is the same
+mechanism working at a smaller scale. Fermata records that a file is no longer
+where it was, and keeps the score, its tags, its practice history and any
+hand-corrected transcription exactly as they were. Put the file back — under the
+old name or a new one — and the next scan clears the mark by itself. Nothing is
+ever deleted because a file went away.
 
 If the files do show up in that listing but not in Fermata, the extension may
 not be one Fermata recognizes. It picks up `.pdf`, `.musicxml`, `.mxl`,
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -207,6 +207,16 @@ restore a backup taken before it.
 That is the message telling you a `git` rollback alone is not enough. Either go
 forward again, or restore the `config/` backup you took before upgrading.
 
+**The version that introduced missing-file tracking is one of those**, and it is
+worth knowing why rather than only that. Older versions deleted a score row
+whenever a scan did not find its file — taking the practice history, tags and
+transcriptions attached to it. Newer ones mark the row instead. An older version
+run against a newer database would not fail on the column it does not know
+about: it would ignore it, read every marked score as though its file should be
+there, not find it, and delete the lot. So the refusal above is not bureaucracy,
+it is the thing standing between a rollback and your practice history. If you
+need to go back, restore the `config/` backup as well.
+
 ## Current limitations
 
 This section is here so you can decide whether Fermata is ready for what you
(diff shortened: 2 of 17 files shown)


sirged:
## Update: both review rounds addressed

Thirteen findings across two reviews. All addressed; the four verified claims left alone. The guard around the design was the problem — it wrote before it decided, could not see the case it existed for, could not be escaped, and was invisible. Those turned out to be one problem.

### Blockers

**1. Rollback destroyed everything.** `SCHEMA_VERSION` → 4, with no migration step behind it, which is new here and is the point: the stamp is what makes an older release refuse. Verified both ways in containers against the real merge-base image:

| | old image starts? | scores | session w/ score | tag links | edited transcriptions |
|---|---|---|---|---|---|
| with the bump | **refuses** ("schema version 4 … understands 3") | 12 | 1 | 1 | 1 |
| stamp forced to 3 | starts | **0** | **0** | **0** | **0** |

Startup now logs its explanation *before* the traceback, since a stack trace is what sends an operator reaching for the old tag. `docs/deployment.md`'s promise is true again, and its Upgrading section now says why this release is one of those.

**2. The refusal was a fixed point.** A refusal now carries a token naming the exact set of unmatched paths; `POST /api/scan/acknowledge` runs the pass with the guard stood down for that evidence and nothing else. Stale consent does not apply (409). Acknowledging also moves the high-water mark down, so a smaller library stops being asked about. Verified in a container: wrong token → 409, right token → 12 marked, tag + 2400s + transcription intact, and the next scan is quiet.

**3. A refused scan was not a no-op.** The decision now reads the directory listing and the stored paths *before* the file loop. Live: `refused: true, added: 0, updated: 0, missing: 0`, and the row set byte-identical. This also closes review 2's blocker — the wrong-folder case (12 rows, 13 unrelated files) is now `account for 0 of the 12` rather than silence.

**4. It was all invisible.** `Library.svelte` renders the refusal, its reason, the paths it could not find, and the confirm control; a missing score is flagged and dimmed on its card; the sidebar shows per-collection missing counts. Two new browser tests drive it end to end.

### Also fixed

- **The ladder.** Collapsed to **one** proportional test, on what would *remain*, against the high-water mark. A mutation loosening the old single-pass boundary changed no outcome — which exposed that test as provably redundant (the high-water mark is never smaller than the current count), so two tests were covering for each other. I proved the redundancy exhaustively before deleting it, and proved that counting *absence* rather than *remainder* reintroduced the fixed point for any library above the floor (my own test had only covered below it).
- **Duplicates** counts files, not marks. **Collections** reports present and missing separately instead of standing a ghost collection with a full count. **Needs attention** excludes a score whose file is gone.
- **Upload** no longer recreates the library root on its way to a subfolder.
- **A content-hash relink is no longer counted as a restore** — that counter is offered as evidence a remount recovered, and only a file returning to its own path supports it.
- **A pass that both marks and adds** warns that the history is on the row nobody will open, instead of reading as clean.
- **`scores` now has a shared column definition** like `practice_sessions`. The column stays in `COLUMN_ADDITIONS` (verified correct); without the shared definition a future rebuild from the schema text would have dropped `missing_since` and turned every missing score back into a present one.
- **Corrected the comment, not the decision:** a returning file finds its transcription again after a plain rename but **not** after an edit-and-move — the hash matches nothing, so the work stays on a row that matches no file. Preserved, better than deleted, permanently detached, and now said so in `db.py` and in the deployment guide.

### Verification

585 server tests (was 528 on main), 148 browser tests, frontend build, container image — all green. **49 mutations applied, 49 killed, 0 survived**, across two harnesses. The mutation run is what found the redundant test and the reintroduced fixed point; two survivors on the first pass led directly to the simplification above.

One branch (`unmatched == 0`) is **unreachable by driving the scanner**, and rather than leave that as a silent gap I said so in the code and tested `_implausible` at the function boundary, where the state is constructible. A system-level test of it could not fail.

### Still open, honestly

- **No way to remove a missing row.** Still no `DELETE /scores`; #56's territory. Less painful than before — such a row is now visibly missing, excluded from "Needs attention", and counted separately in the sidebar — but it is still a row you cannot get rid of.
- **Edit-and-move leaves two scores for one piece.** Nothing lost, permanently detached, documented in the guide.
- **A library below the ten-row floor can still be drained a score at a time** — every step a mark, one good scan restores it, and the last file is caught categorically.
- **`web/**` changes:** `Library.svelte`, `api.js`, `playwright.config.js`, `minimum-tests.js`, `viewer-practice.spec.js`, and one new spec. None touch the metronome.

sirged:
Rebased onto main after #98 (the general metronome) landed. Only `web/tests/minimum-tests.js` conflicted, and it needed adding rather than choosing — #98 raised the browser floor by 37, this branch by 2, so it is 185 and both explanations are kept. `README.md` and `Library.svelte` merged cleanly; the sidebar's metronome link and this branch's refusal panel, missing-file card flag and per-collection missing count all coexist, verified by running the whole browser suite on the merged tree.

All four checks green on the merge: **565 server tests**, **185 browser tests**, frontend build, container image. Locally with `FERMATA_TEST_LIBRARY` set the server suite is 585 passing (CI skips the 20 real-library extraction tests, and says so).

PR #108, opened 2026-08-22
Measure a glyph's position from its ink, not its metrics box (#88)

### LINKED ISSUE #88: Glyph vertical positions are baseline-relative, not ink centres — and the half/whole rest depends on it
Every decoded glyph's vertical centre is a fixed offset away from where the ink actually is, and several decisions compare that value against absolute staff geometry as though it were the ink centre.

## The root cause

The text-trace character box PyMuPDF reports is **metrics**-based, not ink-based: its top and bottom come from the font's ascender and descender relative to the text baseline. So each glyph's recorded centre sits at a constant offset from the baseline rather than at the middle of the drawn shape — measured at **−0.39 staff spaces for Maestro** and **−0.50 for Opus**.

Glyph-to-glyph comparisons are unaffected, because the bias is constant and cancels. Every comparison against *staff* geometry inherits it.

## What it breaks

**The half rest and the whole rest are the same glyph.** Maestro draws both with one glyph — 252 of 296 rest glyphs in the library are that single id — so geometry is the *only* discriminator, and the current rule picks by which staff line the glyph centre is nearest. With the centre offset by nearly half a space, that rule is deciding a **twofold duration difference** from a biased measurement.

The real discriminator is different in kind: a half rest sits *on* a line, a whole rest *hangs below* one. That is sub-space parity against the line grid, not proximity to the nearest line.

Measured consequence on `main`, over 104,448 emitted notes: **3 dotted whole rests** are written out. That is six quarter notes of silence in a library which is almost entirely 3/4 and 4/4 — arithmetically impossible in either. They are not rare curiosities; they are the visible tip of the misclassification, alongside dotted half rests read as dotted wholes and whole rests below the staff read as halves.

**The time signature's numerator/denominator split** uses the same kind of comparison — which digits sit above the middle line and which below — and inherits the same bias.

**The staff-band filters** that decide which glyphs belong to which staff compare the same value against staff extents.

## Why fixing it at the source is the right move

Each of these could be patched at its call site with an offset, but the bias has one cause and one correct fix: **derive a true ink centre once, where the glyph event is created**, and let every consumer compare like with like. That is a smaller change than three or four compensations, and it removes a trap that will otherwise catch the next comparison someone adds.

The font-specific offsets above are measurable per font, and the calibration tables already key on font identity, so the machinery to do it exists.

## Also here

`min(range(len(line_ys)))` raises on an empty line list, while the spacing helper two lines above deliberately tolerates the same case — so a staff with no detected lines throws instead of degrading.

## Note on provenance

Found by review of the glyph-decoding work after it merged, and re-verified against `main` rather than the branch it was written on. The calibration tables themselves are sound — the mean unrecognised-glyph ratio across 2,981 staves is 0.0002 and the font fingerprint holds. These defects are in the **geometry layer that interprets** those tables, not in the tables.

Related: #84 (unmapped glyphs), #85 (invented silence hides losses like these), #86 (absolute distances against scaling geometry).

sirged:
Closes #88.

## The defect

Every glyph's vertical position came from the box `page.get_texttrace()`
reports, and vertically that box is **metrics**-based: its top and bottom are
the font's ascender and descender, so it says the same thing about every glyph
in the font (4.0 staff spaces tall for Maestro, notehead and clef alike). Its
centre therefore sat a fixed distance from the text **baseline** rather than at
the middle of the drawn shape. Measured over the library:

| font | box centre - baseline | glyphs |
| --- | --- | --- |
| Maestro | -0.392 spaces | 123,930 |
| Opus | -0.494 | 10,935 |
| MScore | -0.270 | 10,615 |
| OpusSpecial | -0.668 | 820 |
| Leland (the fixtures' font) | +0.000 | 475 |

Between two glyphs the bias cancels, which is why every duration still decoded.
Against staff geometry it does not - and the half/whole rest decision rested
entirely on it, because Maestro and Opus draw both rests with **one glyph**
(294 of the library's 2,795 rest glyphs are that id).

Note the last row: Leland centres its box on the baseline, so on a fixture page
the box centre and a notehead's ink centre coincide and this defect is
invisible. That is why it had to be found on the library, and why the test that
measures it directly is library-gated.

**Horizontally the same thing was true, in one font**, and that was found while
reviewing this branch rather than before it. `x0`/`x1` come from the advance
width, and a font's two side bearings need not match. Measured over the
library's noteheads:

| font | left side bearing | right side bearing | noteheads |
| --- | --- | --- | --- |
| Maestro | -0.008 spaces | -0.012 | 97,489 |
| MScore | +0.000 | -0.004 | 8,254 |
| **Opus** | **+0.000** | **+0.324** | 8,917 |

A notehead's stem attaches at one side or the other, so on an Opus page "how
far is this stem from the notehead" came out a third of a space larger for an
up-stem, at the right edge, than for a down-stem at the left. That is the whole
margin the choice between two voices' stems was being decided on.

## The fix

`GlyphEvent` now carries the text baseline it was drawn on and its measured ink
extent, and `yc` is the ink centre. The ink comes from the per-glyph bounding
box in the embedded font's `glyf` table - the same bytes the Maestro fingerprint
already hashes, so no outline parsing, and one parse per font resource per
document. Availability: 100% of Opus, OpusSpecial, MScore and Leland glyphs and
126,523 of 126,560 Maestro glyphs; every one of the library's 294 one-glyph
rests. Where it cannot be read the position falls back to the baseline, which is
within 0.013 spaces of a notehead's ink centre and 0.036 of a meter digit's.

That same `glyf` header carries `xMin` and `xMax` beside the `yMin` and `yMax`
it was already unpacking, so **the edges a stem attaches at now come from the
outline too** (`GlyphEvent.stem_edges`), falling back to the metrics box where
the outline cannot be read. The two axes are refused independently: a glyph
drawn as a single vertical stroke has a degenerate x extent and a perfectly good
y one, and folding them into one usability test would cost the vertical reading
the half/whole rest turns on.

Measured against the ink itself once `stem_edges` reads it, the win is real but
not even across fonts. An Opus up-stem's edge distance - the case that
motivated this - falls from 0.348 to **0.024** staff spaces. Maestro's own
side bearings run slightly negative rather than zero, though, so on Maestro -
**95% of the library's noteheads** - the ink edge is 0.004 to 0.008 spaces
*worse* than the metrics box was. The correction is still the right one - it
measures from the side a stem actually attaches at rather than a side bearing
that happens to cancel on most fonts - but the Maestro cost is real and stated
here rather than left out.

Four readings change with it.

**The half/whole rest** is now sub-space **parity** against the line grid. A
half rest sits *on* a line and a whole rest *hangs below* one, and both are half
a space deep, so each one's ink centre lands a quarter space off the grid -
above for a half, below for a whole. That is a different kind of measurement
from "which line is it nearest", not a tighter one: it holds at every **line**
of the grid, including the ledger positions a second voice's rests use, where
nearest-line read every whole rest below the middle line as a half.

It does **not** hold for an arbitrary displacement, and the earlier wording of
this section claimed it did. Parity is measured modulo one space, so displacing
a rest by an odd number of half spaces lands it exactly where the other rest
would sit and returns that other rest with `decided=True` - a twofold duration
error reported as a reading. Nothing in `(yc, line_ys, spacing)` can separate
the two cases, so this is now stated in the docstring and pinned by a test
rather than guarded. It has no current effect: all 256 one-glyph rests the
decode reads land between 0.235 and 0.266 of a space from a line, and every one
is decided.

Checked against the only population where the answer is known *without*
geometry - the rests a SMuFL font names in the codepoint itself: parity agrees
with the engraving on all 20 in the library, nearest-line on 16.

**A flag counts when the stem's tip is inside its ink**, not near its centre. A
32nd hook's ink centre is two spaces from the tip it hangs off; the fixture's
four 32nds decoded as quarters once positions became ink-based, and a
centre-distance tolerance wide enough to cover them would reach four spaces to
either side - far enough to take a neighbouring voice's hook. The tip is inside
the ink for 4,669 of the library's 4,843 attachments and the rest fall within
the 0.25-space pad, whose value sits on a plateau (0.1 and 0.25 count the same
5,158 hooks; no pad at all loses 166).

**An augmentation dot** goes to the note whose relation to it an engraver would
draw - the note in the dot's own space, then the note on the line below it, then
the note on the line above - rather than to the nearest notehead. A dot is
always in the middle of a space (11,350 of 11,405 in the library, within a tenth
of a space), so two noteheads a space apart are exactly equidistant from the dot
between them and nearest-distance decided that tie on rounding. It gave one
notehead two *vertically stacked* dots - which is not what a double dot is; a
double dot is two dots side by side - while the note the second dot belonged to
lost half its length. Both engravings the ranking has to serve are in the
library: a chord of noteheads a space apart each with its own raised dot, and a
lower voice's dotted whole note whose dot is in the space below it.

**Which stem a notehead hangs off** is now ranked on both distances together.
This is the defect the ink centres exposed and it is described in full in the
next section, because it was a new one rather than a consequence of better
voice grouping.

Also here, from the issue: an empty staff-line list no longer raises, and a rest
whose value was guessed rather than read caps what the staff may claim instead
of sitting unread in the decode stats.

## The stem-attachment defect this branch introduced, and its fix

`_best_stem` ranked candidate stems on `x`, using `y` only to break `x`'s ties.
Putting a notehead at its true vertical position brought a neighbouring voice's
stem end inside the y gate, where it then won on a hair of x.

In the commonest two-voice figure in this repertoire - a melody note sharing a
beat with a stem-down chord - **both voices' stems are at the same place
horizontally**, one at each side of the notehead. Their x distances therefore
differ by the width of a stem stroke and their y distances by a whole staff
space, and x was deciding. Measured over the library's 98,737 notehead-to-stem
lookups, against `main`:

| | main | ink centres only | with the fix |
| --- | --- | --- | --- |
| took a stem further away in y than an in-range alternative | 0 | **90** (20 scores) | 7 (6 scores) |
| took a stem >0.8 spaces off in y with one <=0.3 available | 1 | **93** | 2 |
| ...of those, the other voice's stem | 1 | **49** | **0** |

Two verified reproductions, both checked against the printed page:

- **Dalza's Recercar**, score measure 11: a filled eighth with its own up-stem
  one space above a stem-down half-note chord. `main` emits two voices of four
  quarters each in the 2/2 bar, correctly. With ink centres it emitted a single
  voice of six half-note units and no second voice - a fourfold duration error
  with a voice lost. Measures 7, 12, 13 and 14 the same way. This is also the
  Opus side-bearing case: against the metrics box the note's own up-stem
  measured 1.75pt away and the chord's down-stem 0.16pt.
- **Easy Christmas Songs Vol 1-4**, page 60 bar 322: the beat-3 melody quarter
  was absorbed into the bass half chord, leaving voice 1 three quarters long.
  The chord's down-stem won by 0.16pt of x over a stem 6.2 times closer in y.
  Bar 324 lost voice 2 entirely. Also reproduces at bars 326, 334, 401 and 405.

Candidates are now ranked on both distances together, each expressed as a
fraction of the slack its own axis is already allowed - no third constant to
tune. `y` cannot rank them alone either: a neighbouring note's stem a
notehead-width away can end nearer this notehead's centre than its own stem
does, and 20 noteheads go to such a stem if x is ignored, so both a y-first
ranking and the existing x-first one are wrong in opposite directions.

**A stem-direction consistency check inside `_best_stem` was tried and is not
better.** The check `_assign_stem_directions` already makes - an up-stem leaves
a notehead at its right edge, a down-stem at its left - needs the mean x of
every notehead on the stem, which one candidate notehead does not have. It was
tried anyway. Two independent attempts to measure how many noteheads it leaves
with no consistent candidate at all, and how many selections it changes,
disagreed with each other, so neither count is published here - but the
selections it does change are not better: some individual noteheads move
closer to their true stem and others move away, and the change is not an
improvement on net. That structural reason, which both measurements agree on,
is now recorded in the function's docstring so it is not tried again.

The 7 remaining noteheads that take a stem further in y than `main` chose were
each looked at individually. All 7 are the same shape and the new choice is
right in every one: the note's own up-stem sits on its right ink edge, 0.008
spaces away, while the stem `main` preferred is an up-stem 0.55 to 0.67 spaces
clear of the notehead's *left* edge - a neighbouring note's, which an up-stem on
the left cannot be anything else. Both remaining rows of the second metric are
the same kind of correction.

## What moved, across all 297 scores

| | main | ink centres only | with the stem fix |
| --- | --- | --- | --- |
| **dotted whole rests** | **3** | **0** | **0** |
| defective bars | 6,411 | 5,931 | **5,863** |
| overfull bars | 2,335 | 1,943 | **1,896** |
| short bars | 4,668 | 4,440 | **4,435** |
| bars holding invented silence | 3,980 | 3,759 | 3,753 |
| invented silence (quarters) | 5,831.6 | 5,450.5 | 5,529.2 |
| notes emitted | 98,688 | 98,688 | **98,688** |
| `<chord>` markers | 17,930 | 17,279 | 17,231 |
| rhythm confidence | 15 high | 26 high | 27 high |
| meter / key / fret confidence | unchanged | unchanged | unchanged |

Pitched-note counts are identical in all three columns: nothing was dropped or
added, only re-read. `<note>` elements are not quite: over the 35 scores this
branch touches, 16,444 become **16,445**, and `<rest/>` elements 229 become
**230** - one extra rest, not a note.

The `<chord>` count falls in both steps, and the direction is the point rather
than a side effect. A `<chord>` marker is a notehead emitted as a member of the
beat before it, so a melody note wrongly attached to the accompaniment's stem
*adds* one. Removing 48 of them is 48 noteheads that stopped being folded into a
chord they are not in - the Christmas bar above is one of them.

Invented silence rises slightly against the ink-centre column while the bars
holding it fall. That is consistent: separating two voices that were being read
as one means both voices now have to be padded to the bar, where before there
was one voice to pad.

Per file against the ink-centre column: 15 scores improve by 71 defective bars
between them, and 2 lose 3 between them - "Score A" (34 -> 36) and "Score B" (22 -> 23) - with the rest unchanged.
The two that lose ground were not individually diagnosed; in the larger of them
the stem the fix now chooses is the one touching the notehead's edge, so the
bar was previously balanced by a compensating error rather than read
correctly.

## Blast radius, stated rather than netted

The previous version of this section gave a net figure. The gross figures,
measured at two levels, against `main`:

At the **stem-end lookup** (`_best_stem` itself), of 98,737 noteheads:

| | ink centres only | with the stem fix |
| --- | --- | --- |
| lose a stem end | 1,543 | 1,543 |
| gain a stem end | 1,279 | 1,279 |
| attach to a *different* stem end | 1,905 | 2,922 |
| **any change** | **4,727 (4.8%), 270 of 297 scores** | **5,744 (5.8%), 270 scores** |

At the **final attachment**, after the threading pass and the direction check,
of 101,073 noteheads:

| | ink centres only | with the stem fix |
| --- | --- | --- |
| lose a stem | 164 | 147 |
| gain a stem | 215 | 219 |
| attach to a different stem | 1,873 | 2,899 |
| stem **direction** flips | 2,238 | 3,239 |
| **any change** | **2,252 (2.2%), 241 scores** | **3,265 (3.2%), 243 scores** |

The earlier text said 292 fewer noteheads find a stem end. The measured figure
is **264** (11,654 to 11,918), of which the threading pass picks up **150**, so
**114 more** find no stem at either stage. That is not the end of it, and the
earlier text stopped too early: `_assign_stem_directions` then drops stems whose
side and overhang contradict each other, and it drops 358 at `main` against 172
here, so the number of noteheads that finally carry no stem falls from 2,892 to
2,820 (excluding whole notes, which never take one by definition). The stem fix
leaves all of these figures unchanged - it changes *which* stem is chosen, not
whether one is found.

**What a notehead with no stem is given.** Not a whole note. It keeps the value
its shape implies with no flag or beam counted, so a filled head is emitted as a
plain quarter (1,954 of them), a half note as a half, a diamond as a quarter,
and it loses its voice signal, which folds it into whatever else sounds at its
onset. The cost is therefore an eighth or a sixteenth read *long* as a quarter,
plus the voice - never a note lengthened to a whole. A second population, 172 of
them, found a stem and had its flags counted before the direction check dropped
it; those keep their duration and lose only the voice.

## One-glyph rests

The earlier text said 139 of the library's one-glyph rests move from whole to
half and 10 the other way. Measured at the decode boundary - the value
`decode_note_events` gives a `rest_half_whole` glyph, over the 256 such rests
the decode reads, matched one-to-one between the two trees:

- **140** move whole to half, removing 280 quarters of silence that was never
  printed. This is why overfull bars fall furthest.
- **12** move half to whole, the permissive direction, adding **24** quarters.

Both figures are the same with and without the stem fix. A review of this branch
put the permissive count at 26; that could not be reproduced at this measurement
point, and the definition used here is stated above so it can be checked.

**The numerator/denominator split did not move**, and that is measured rather
than assumed: every time-signature digit in the library sits exactly +/-1.0
space from the middle line, so a 0.39-0.49 space bias never crossed it. It did
eat most of the denominator's margin - down to 0.51 spaces for Opus - and would
have crossed on a condensed or large-print meter. The staff-band filters
likewise never misfiled a glyph at this scale. Both are now measured against the
ink for the same reason.

## Other invariants, once the first one was clean

The issue's technique applied further. Counting what the emitted MusicXML can
never legally contain:

- **Dotted whole rests: 3 -> 0.**
- **Rests longer than their bar: 7 -> 1.** The one that remains is a whole rest
  alone in a 3/4 bar - not a defect in the engraving but in the reading: a
  whole-measure rest is drawn as a whole rest whatever the meter, and this
  decoder reads it as four quarters. Worth its own issue.
- **Dotted whole notes: unchanged.** 60 of them by the count used here (every
  `<note>` carrying `<type>whole</type>` and a `<dot>`), and the same 60 on
  `main` and on both commits of this branch. Most are in 12/8, where six
  quarters is exactly the bar and the reading is right; the remainder are in
  bars read as 4/4, where six quarters is impossible and either the meter or
  the dot is wrong. Untouched by this change; worth its own issue.
- Notes longer than their bar: 327 on `main`, 332 with ink centres, **331** with
  the stem fix. Scores that gain a dot the old tie-break dropped were already at
  or over their bar on that note.

## Documentation

`docs/musicxml-tab-profile.md`'s Rule 14 paragraph quotes library figures for
the MuseScore round trip, and both commits on this branch move them. The
experiment was re-run at the head of the branch (MuseScore 4.6.3, the Rule 8
check that section describes implemented independently of the extractor, all 293
emitted files): 6,155 `<forward>` elements written, 2,957 surviving a re-save as
one, and **not one of them still carrying the marking**; defective measures
5,863 before the save against 5,430 after it, so **433** measures read as
conforming downstream of a save. The independent checker was validated by
reproducing the previous commit's published figures exactly (5,931 -> 5,600,
5,450.5 -> 3,848.2 quarters, 3,109 `<forward>`) before being used for the new
ones.

No fixture's expected values changed. One test assertion was renamed
(`dot_y_tol` -> `dot_x_tol` in the tolerance-scaling test) because a dot's
vertical window is no longer a symmetric tolerance.

## Tests, and proof they can fail

26 new tests. From the first commit: the parity rule at every line of the grid
and above and below the staff, the undecided cases, the decode reading both
rests from the one glyph, the SMuFL oracle on the committed fixture, flag
containment in both directions and its fallback, the dot ranking's two real
engravings, the degraded-provenance report, and a library-gated measurement that
every notehead's ink centre lands on the staff's own half-space grid (worst
0.020 spaces) while the box centre it replaced never does (best 0.098).

Nine more for the stem fix and the corrections: both reproductions at the
geometry the page really has, the Opus side bearing, the metrics-box fallback,
the y-first failure mode, the ink reader's per-axis refusal, the horizontal
placement's direction, an end-to-end assertion that Dalza's five two-voice bars
keep both voices, the parity rule's confidently-wrong half-space case, and the
`_Tol` spacing floor.

**One test on this branch could not fail and has been replaced.** It asserted
`half_or_whole_rest`'s guard against a zero or negative spacing, which nothing
can reach: `_Tol` floors a missing, zero or negative spacing to the reference
value before the rest reading ever sees it. The floor is now what is pinned, and
the reachable half of what that test claimed - a staff whose lines were not
detected - now goes through `decode_note_events` instead of calling the helper
directly.

Every new test was proven to fail by breaking the behaviour it describes.
Twenty mutations, each caught; the eight for this commit:

| mutation | tests turned red |
| --- | --- |
| ranking back to `(dx, dy)` lexicographic | 2, incl. the end-to-end Dalza bars |
| ranking changed to `(dy, dx)`, y first | 1 |
| `stem_edges` back to the metrics box | 1 |
| `stem_edges` loses its box fallback | 1 |
| horizontal ink flipped like the vertical one | 11, incl. 4 committed fixtures |
| ink reader uses one usability test for both axes | 1 |
| `half_or_whole_rest` loses its empty-line guard | 1 |
| `_Tol` no longer floors a non-positive spacing | 2 |

Server suite: 630 passed, 8 skipped with the library set; 606 passed, 32 skipped
without it, which is the CI-equivalent run.

## Not verified

- The rotated-span guard (17 characters in the library are rotated, none of them
  a music glyph on a notation staff) and the CFF path have no test: the fixtures
  cannot produce either, and both only ever degrade to the baseline.
- Bravura's 10 glyphs on one library page get no ink box, because its name
  resolves to more than one embedded resource on that page and a box could then
  come from the wrong glyph order. They are unrecognised harmonic noteheads that
  only feed the honesty counters.
- The three scores that lose defective bars to the stem fix were counted, not
  diagnosed bar by bar.
- The permissive-direction rest count could not be reconciled with a review's
  figure of 26; 12 is what this measurement point gives.
- The remaining whole-bar rest and the 17 impossible dotted whole notes are
  named above, not fixed.

### DIFF
--- a/docs/musicxml-tab-profile.md
+++ b/docs/musicxml-tab-profile.md
@@ -478,10 +478,9 @@ that needs it has to get it from the producer.
   measures fall from 5,931 to 5,600 and identifiable inferred silence from
   5,450.5 to 3,848.2 quarter notes, so **331 measures this profile reports
   defective read as conforming to anything downstream of that save**. The
-  agreement between a
-  producer's figures and an independent check holds for the file *as the
-  producer wrote it*. Verify against that file, not against a round trip
-  through an editor — see [Checking a file](#checking-a-file).
+  agreement between a producer's figures and an independent check holds for the
+  file *as the producer wrote it*. Verify against that file, not against a
+  round trip through an editor — see [Checking a file](#checking-a-file).
 
 **In the other direction.** Fermata also emits the same music as alphaTex for
 its transcription editor. That format has no editorial mechanism for this and
--- a/docs/musicxml-tab-profile.md
+++ b/docs/musicxml-tab-profile.md
@@ -469,18 +469,20 @@ that needs it has to get it from the producer.
   the voice there; MuseScore rewrites it as an invisible rest. A `<forward>`
   anywhere else must advance the position, or every note after it sounds early.
 - **The marking does not survive a save by another program.** Measured over the
-  same library with MuseScore 4, re-saving the 293 files it emits: every one of
-  the 3,109 `<forward>` elements that survives as one loses its `<footnote>`
-  and its `<voice>`, a trailing one is rewritten as an invisible rest and a
-  leading one as a `<backup>` with no `<forward>` at all. Net effect on a Rule
-  8 check of the re-saved files — a measure counted once whichever way it is
-  wrong, which is the `bars_defective` figure below — is that defective
-  measures fall from 5,931 to 5,600 and identifiable inferred silence from
-  5,450.5 to 3,848.2 quarter notes, so **331 measures this profile reports
-  defective read as conforming to anything downstream of that save**. The
-  agreement between a producer's figures and an independent check holds for the
-  file *as the producer wrote it*. Verify against that file, not against a
-  round trip through an editor — see [Checking a file](#checking-a-file).
+  same library with MuseScore 4, re-saving the 293 files it emits: of the 6,155
+  `<forward>` elements written, 2,957 survive as one and every one of those
+  loses its `<footnote>` and its `<voice>`, while a trailing one is rewritten as
+  an invisible rest and a leading one as a `<backup>` with no `<forward>` at
+  all. Not one `<forward>` in the re-saved files still carries the marking. Net
+  effect on a Rule 8 check of the re-saved files — a measure counted once
+  whichever way it is wrong, which is the `bars_defective` figure below — is
+  that defective measures fall from 5,863 to 5,430 and identifiable inferred
+  silence from 5,529.2 quarter notes to none that can be identified at all, so
+  **433 measures this profile reports defective read as conforming to anything
+  downstream of that save**. The agreement between a producer's figures and an
+  independent check holds for the file *as the producer wrote it*. Verify
+  against that file, not against a round trip through an editor — see
+  [Checking a file](#checking-a-file).
 
 **In the other direction.** Fermata also emits the same music as alphaTex for
 its transcription editor. That format has no editorial mechanism for this and
--- a/docs/musicxml-tab-profile.md
+++ b/docs/musicxml-tab-profile.md
@@ -469,13 +469,16 @@ that needs it has to get it from the producer.
   the voice there; MuseScore rewrites it as an invisible rest. A `<forward>`
   anywhere else must advance the position, or every note after it sounds early.
 - **The marking does not survive a save by another program.** Measured over the
-  same library with MuseScore 4: every one of 3,464 `<forward>` elements loses
-  its `<footnote>` and its `<voice>`, a trailing one is rewritten as an
-  invisible rest and a leading one as a `<backup>` with no `<forward>` at all.
-  Net effect on a Rule 8 check of the re-saved files: defective measures fall
-  from 6,013 to 5,707 and identifiable inferred silence from 5,831.6 to 4,466.2
-  quarter notes, so **306 measures this profile reports defective read as
-  conforming to anything downstream of that save**. The agreement between a
+  same library with MuseScore 4, re-saving the 293 files it emits: every one of
+  the 3,109 `<forward>` elements that survives as one loses its `<footnote>`
+  and its `<voice>`, a trailing one is rewritten as an invisible rest and a
+  leading one as a `<backup>` with no `<forward>` at all. Net effect on a Rule
+  8 check of the re-saved files — a measure counted once whichever way it is
+  wrong, which is the `bars_defective` figure below — is that defective
+  measures fall from 5,931 to 5,600 and identifiable inferred silence from
+  5,450.5 to 3,848.2 quarter notes, so **331 measures this profile reports
+  defective read as conforming to anything downstream of that save**. The
+  agreement between a
   producer's figures and an independent check holds for the file *as the
   producer wrote it*. Verify against that file, not against a round trip
   through an editor — see [Checking a file](#checking-a-file).
--- a/server/fermata/glyph_rhythm.py
+++ b/server/fermata/glyph_rhythm.py
@@ -1611,10 +1611,22 @@ def half_or_whole_rest(yc, line_ys, spacing):
     The discriminator is sub-space PARITY against the line grid - which side
     of a line the ink sits on - and not, as this used to ask, which staff line
     the glyph is NEAREST. The difference is not a tolerance, it is a kind:
-      * Parity holds wherever on the staff the rest was engraved, including
-        the ledger positions above and below it that a second voice's rests
-        use. Nearest-line put every rest below the middle line in the same
-        bucket, so a whole rest hanging under the staff read as a half.
+      * Parity holds at every LINE of the grid, including the ledger positions
+        above and below the staff that a second voice's rests use. Nearest-line
+        put every rest below the middle line in the same bucket, so a whole
+        rest hanging under the staff read as a half.
+
+        It does NOT hold for an arbitrary displacement, and this is a real
+        limit rather than a tolerance: the signal is which quarter of a space
+        the ink sits in, measured modulo one space, so displacing a rest by an
+        ODD number of half spaces lands it exactly where the other rest would
+        be and returns that other answer with decided=True. Nothing in (yc,
+        line_ys, spacing) can separate the two cases - a half rest sits a
+        quarter space above a line and a whole one a quarter space below one,
+        and a half-space shift maps each onto the other - so this is stated,
+        not guarded. No engraving in the library needs the guard: all 256
+        one-glyph rests the decode reads land between 0.235 and 0.266 of a
+        space from a line, and every one is decided.
       * Parity is measured against the ink. Nearest-line was measured against
         the metrics box centre (see GlyphEvent), which is a fixed distance
         from the BASELINE - nearly half a space, most of the way to the other
--- a/server/fermata/glyph_rhythm.py
+++ b/server/fermata/glyph_rhythm.py
@@ -648,9 +648,14 @@ class _InkBoxes:
         (which is most of them - a Maestro subset keeps all 204 slots and
         fills the handful a page uses).
 
-        Whether the box is USABLE is asked per axis by span and xspan, because
-        the two axes fail independently: a glyph drawn as a single vertical
-        stroke has a degenerate x extent and a perfectly good y one.
+        Whether the box is USABLE is asked per axis by span and xspan, and the
+        two axes are refused independently BY CONSTRUCTION rather than
+        because either has been seen to fail alone: a glyph drawn as a single
+        vertical stroke could in principle have a degenerate x extent and a
+        perfectly good y one. No glyph in the library actually splits that
+        way - 0 of 1,855,076 header reads have one axis usable and the other
+        not - but asking per axis costs nothing and is the right shape for a
+        case the library just hasn't shown yet.
         """
         if gid in self._cache:
             return self._cache[gid]
@@ -1725,7 +1730,7 @@ def _best_stem(stems, stem_xs, x0, x1, yc, tol, x_tol=None, y_tol=None):
     (0.348 at the 99th percentile) and its end sits 0.175 spaces from the ink
     centre (0.179 at the 90th). Both are sharp; neither decides alone, and
     letting the noisier margin overrule the other picked a stem further away in
-    y than the alternative for 51 noteheads across 20 scores, in 49 cases the
+    y than the alternative for 90 noteheads across 20 scores, in 49 cases the
     other VOICE's stem. That is a duration error of up to fourfold and a lost
     voice, not a near miss - see the two-voice bars of Dalza's Recercar.
 
@@ -1756,9 +1761,11 @@ def _best_stem(stems, stem_xs, x0, x1, yc, tol, x_tol=None, y_tol=None):
             continue
         dx = min(abs(s.x - x0), abs(s.x - x1))
         # Rounded so two candidates a floating-point hair apart cannot swap on
-        # the platform's last bit, and tie-broken on y: two stems the same
-        # distance away are two voices, and the one whose end lands on this
-        # notehead is the one it hangs off.
+        # the platform's last bit. The second key element exists to make the
+        # ordering TOTAL rather than merely partial - worth keeping even
+        # though it has not once had to decide anything: 0 exact ties on the
+        # rounded hypot across 6,203 multi-candidate noteheads in the
+        # library, and deleting it changes nothing measured.
         key = (round(math.hypot(dx / xt, dy / yt), 6), round(dy, 3))
         if best_key is None or key < best_key:
             best, best_key = s, key
@@ -1911,9 +1918,14 @@ def _assign_stem_directions(notes, stems_by_key):
     This check is a net, not a ranking, and it cannot be moved into
     _best_stem to do the choosing: the side test needs the MEAN x of every
     notehead on the stem, which is exactly what one candidate notehead does
-    not have. Applied per candidate it rejects the right stem too - measured
-    over the library, it leaves 344 noteheads with no consistent candidate at
-    all and changes 38 selections, and the selections it changes are wrong.
+    not have. It was tried anyway. Two independent attempts to measure how
+    many noteheads it leaves with no consistent candidate at all, and how
+    many selections it changes, disagreed with each other, so neither count
+    is published here - but the selections it does change are not better:
+    some individual noteheads move closer to their true stem and others move
+    away, and the change is not an improvement on net. The structural reason
+    above, which both measurements agree on, is why this is not retried
+    rather than the numbers.
 
     This catches only the contradictory subset. A neighbouring voice's stem
     that happens to sit where this notehead's own stem WOULD sit is
@@ -2030,6 +2042,11 @@ def _assign_dots(owners, dot_events, tol):
     counts = collections.Counter()
     if not owners or not dot_events:
         return counts
+    # Still the metrics box's x1 here, not an ink edge - left inconsistent
+    # with the notehead-to-stem lookups above on purpose, since it was
+    # measured against the library and found to have no current effect on
+    # which owner a dot is given, the same way the flag8_or_rest_quarter
+    # window beside it was.
     owners = sorted(owners, key=lambda e: e.x1)
     owner_x1s = [e.x1 for e in owners]
     for dot in dot_events:
@@ -2207,6 +2224,11 @@ def decode_note_events(page, staff_top, staff_bottom, staff_x0, staff_x1, line_y
             # (Barlines are excluded from `stems` - see _is_barline - or a
             # rest engraved next to one would vanish here.)
             if ev.category == "flag8_or_rest_quarter":
+                # Still the metrics box here, not ev.stem_edges - left
+                # inconsistent with the notehead lookups above on purpose,
+                # since it was measured rather than assumed: the ink and box
+                # windows agree on all 67 of the library's flag8_or_rest_quarter
+                # glyphs, so switching this one has no current effect.
                 if _has_stem_near(stems, stem_xs, ev.x0 - tol.rest_stem_pad,
                                   ev.x1 + tol.rest_stem_pad, ev.yc, tol):
                     continue  # it's a flag - counted via the notehead's stem (see FLAG_HOOKS)
--- a/server/fermata/glyph_rhythm.py
+++ b/server/fermata/glyph_rhythm.py
@@ -96,12 +96,23 @@ How this works (full validation detail):
   measured across 254 staves), so a condensed multi-system score or a
   large-print edition silently dropped stems and beams and degraded every
   eighth and sixteenth to a quarter while still reporting high confidence.
+
+  A glyph's vertical position is its INK centre, read from the embedded
+  outline - NOT the centre of the box the text trace reports, which is a
+  metrics box whose top and bottom are the font's ascender and descender and
+  therefore says the same thing about every glyph in the font. See
+  GlyphEvent, _InkBoxes and half_or_whole_rest: the difference is about 0.4
+  staff spaces, it is invisible to glyph-to-glyph comparisons because it
+  cancels, and it was deciding whether a rest was a half or a whole - a
+  twofold difference in duration - from a measurement that was never the
+  rest's position.
 """
 import bisect
 import collections
 import hashlib
 import io
 import math
+import struct
 import weakref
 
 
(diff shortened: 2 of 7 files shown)


sirged:
## Blocking: the stem tolerance was not recalibrated for the new notehead centre

The ink-centre correction itself is verified right — the half/whole rest direction was checked against four printed pages and the previous behaviour was badly wrong, reducing in practice to "whole iff nearest line index <= 1" with the answer sometimes decided by a floating-point tie. That part should land.

What blocks it is a side effect that this PR describes as "a second-order effect of better voice grouping". It is not. It is a new defect.

`_best_stem` ranks candidates on `key = (round(dx, 3), round(dy, 3))` — a tuple, so **x is compared first and y only breaks ties in x**. Moving a notehead down to its true vertical position pushes it into the y-gate (`stem_y_tol`, 1.17 spaces) of a neighbouring voice's stem, which then wins purely on x.

Measured on `Dalza-Recercar-Guitar-2019.pdf` page 1, 4th system, 2nd bar (score measure 11), an unambiguous engraving — a filled eighth one space above a stem-down half-note chord:

```
                            own stem          other voice's stem
box centre  (base)    dx=1.720 dy=1.680   dx=0.150 dy=8.190 (outside y-gate)
ink centre  (this PR) dx=1.720 dy=0.760   dx=0.150 dy=5.750 (inside, by 0.042pt)
```

The own stem's end lands *on* the notehead and it still loses, because 0.150 < 1.720 decides it before y is looked at.

End to end in that measure, the base emits v1=1920 and v2=1920 in a 2/2 bar — exactly right. This branch emits **v1=2880 in a 1920 bar and no v2 at all**: the melody eighth becomes a `half` chord member, a 4x duration error, and voice separation for the bar is lost. Measures 7, 12, 13 and 14 are damaged the same way.

Library-wide, of the 2,043 noteheads that switch stems, **104 across 21 files switch to a stem end that is *further away in y*** — the signature of x-before-y ranking. They cluster in `Classical-Guitar-Method-Vol1-2020.pdf` (27), `Easy-Christmas-Songs-for-Guitar-Vol1-4.pdf` (24) and `Dalza-Recercar-Guitar-2019.pdf` (14) — two of which are scores this PR lists as losing ground.

### The blast radius needs stating, not netting

"292 fewer noteheads find a stem END" is a net of much larger movement:

| | count |
|---|---|
| lose their stem | 1,601 |
| gain a stem | 1,309 |
| switch to a different stem | 2,043 |
| ...of which stem **direction** flips | 2,013 |
| total noteheads affected | 4,953 (5.0% of 98,665) in 269 of 297 scores |

Stem direction is precisely what voice separation keys on, so 2,013 flips are 2,013 potential voice reassignments, and **all 11 scores promoted from low to high confidence have between 1 and 20 direction flips**. Bar arithmetic is the only thing guarding that, and arithmetic cannot see a chord grouping that is wrong but internally consistent.

### What the 651 figure actually was

Sounding notes are **identical at 98,688**. Nothing is dropped and nothing is duplicated. The movement is that 651 noteheads lost their `<chord>` marker (17,930 -> 17,279 across 179 files) and now advance time instead of stacking — a substantive re-grouping of 651 simultaneities that the PR does not quantify.

### Two figures in the description are wrong

Re-running both trees' own decoder over all 297 scores: one-glyph rests moving whole -> half is 141, and half -> whole (the permissive direction, a rest becoming twice as long) is **26, not 10** — a 2.6x understatement, +52 quarters of silence. 11 of the 167 movements are on dotted rests.

### Also not true as stated

"Parity holds wherever on the staff the rest was engraved" holds only where the rest sits on the *line* grid. Displace it an odd number of half-spaces and the rule returns the other rest with `decided=True` — the offset is still about 0.25, so the window guard does not catch it. Confirmed on an engraved fixture. No instance exists in the library (all 294 one-glyph rests fall in 0.232-0.266), so there is no current impact, but the failure mode is a confident twofold duration error rather than an honest `undecided`.

`test_a_staff_with_no_detected_lines_is_undecided_rather_than_a_crash` also tests a **dead branch** for the spacing half: `_spacing_from_lines` returns the reference constant for a 0- or 1-line list and `_Tol` floors to the same value, so `not spacing or spacing <= 0` is unreachable from `decode_note_events`.

### To unblock

Rank on vertical distance, or on a combined distance, rather than x-then-y — and re-measure. The target is falsifiable: **the 104 further-away switches go to 0**, Dalza measure 11 returns to v1=1920 / v2=1920, and the aggregate figures do not regress. Please also state the real blast-radius numbers and correct the 10 to 26.

sirged:
## Second review, independent route, same root cause

A code review reached `_best_stem`'s lexicographic `dx` dominance from the opposite direction — the real library with printed-page verification, rather than a constructed engraving. The two sets of figures agree once units are reconciled (1.17 spaces = 5.79 pt). Three things it adds:

**A different and stricter measurement of the blast radius.** Counting noteheads that chose a stem more than 0.8 spaces away *while an alternative under 0.3 spaces existed*: **1 -> 71**, of which **49 point the opposite way** — cross-voice. Spread over 20 scores, and critically **17 of those scores show no rise in defect count**, so those bars are read wrong and the bar arithmetic never notices. On the worst score 11 bars change, and bar 324 has **voice 2 emptied entirely**. The reported "+5" understates it.

That is a stricter metric than the 104 further-away switches from the other review; they are complementary views of the same phenomenon, not competing numbers.

**Why this slipped through.** `stem_x_tol` and `stem_y_tol` were calibrated against metrics-box centres and were **never re-calibrated for the ink change** — unlike `flag_ink_pad` and `_DOT_Y_SLACK`, which were. Some tolerances were revisited and these two were missed.

**A structural gap worth fixing rather than tuning around.** `_stem_through_notehead`'s docstring reasons correctly about stem direction separating voices, but `_best_stem` — where the confusion actually happens — has **no stem-direction check at all**. Given that 2,013 of the switches are direction flips, that looks closer to the real fix than adjusting a tolerance.

## The explanation in the description does not reproduce

"292 fewer noteheads find a stem END ... picked up by the threading pass" checks out as follows:

- the count is **263**, not 292
- the threading pass catches **150 of 264**, leaving **114 noteheads with no stem at all** — unmentioned in the description
- on the actually-regressed notehead the threading pass is **never reached**: it *does* find a stem end, just the wrong one

So the reasoning that made the regression look like a benign second-order effect describes a path the regression does not take, and the real mechanism was never examined. Those 114 stem-less noteheads need accounting for in the fix as well.

## The confidence rise is earned — with one caveat that outlives this PR

This was the question I most wanted answered, and the answer is favourable. Exactly **11 scores rise and 0 fall**. Four risers were checked against the printed page across 7 bars; in every case the head reading matches the engraving and the base reading did not. The recurring situation is textbook polyphony — a filled up-stemmed melody notehead a space above an open down-stemmed bass chord — which the base wrongly chorded together, giving the melody note the open notehead's duration. In one case that produced 5 quarters in a 4/4 bar.

It is also **not self-suppression**, which was the failure mode I was most worried about. Instrumented over 2,749 decoded staves per revision: `unknown_at_flag_position` = 0 on both sides, `undecided_rests` = 0, and `rhythm_provenance` is **identical** on both — no staff changed provenance anywhere. The label moved only because the defective-bar ratio crossed the 0.25 gate.

The caveat is pre-existing and I have filed it separately: the weakest riser is labelled "high" with **22.9% of its bars not adding up**. That gate is too permissive for a tool whose main claim is honesty about what it could not read.

Issue #117, opened 2026-08-23
A score whose rhythm came from spacing rather than glyphs never says so


sirged:
Found by a self-audit of the honesty surface after #108 merged.

`rhythm_provenance` records how each staff's rhythm was obtained. Measured over the library it is roughly:

```
glyphs           2,656 staves
glyphs-degraded      3
spacing            125
```

Rhythm read from **glyphs** means flags, beams, dots and rest shapes were decoded. Rhythm derived from **spacing** means durations were inferred from the horizontal gaps between noteheads — far weaker, and wrong whenever the engraver's spacing is not proportional, which is common in hand-adjusted scores and anywhere a system is justified to the margin.

The field is computed in `tabextract.py`, carried on `ExtractionResult`, and serialised into the confidence blob. **Nothing tells the user.** There is no warning prose for it — grepping for any message about spacing-derived rhythm finds nothing — and no interface code reads the field. So 125 staves' worth of durations that were guessed from geometry present identically to durations that were read.

This is the same failure this project keeps repeating and has already filed twice: the measurement gets built and verified, and the half that tells the user gets left. See #103 and #115. Worth treating the three together rather than one at a time, since they all want the same surface.

### Verifying a fix

Warning prose naming the count and the affected staves, on the same footing as the existing padded-bars message — which is the model to copy, because that one does reach the interface. A score with any spacing-derived staff should not be able to present as fully read. Confirm against a score known to have one, and confirm the count matches `rhythm_provenance`.

Note while fixing: the interface consumes warnings as a list of strings and loops over them, so prose reaches the user automatically. A new *count field* on its own would not — which is exactly how this got missed.

Issue #114, opened 2026-08-22
A score can be labelled high confidence with nearly a quarter of its bars not adding up


sirged:
Surfaced by the confidence audit on #108. Pre-existing and independent of that PR.

The rhythm confidence label is gated on the fraction of defective bars crossing 0.25. A score therefore reads **high** with up to a quarter of its bars failing to sum to their meter. Measured on the current library, the weakest score carrying a "high" label has **22.9% of its bars not adding up**.

That is the wrong shape for this tool. The entire argument for the extraction pipeline is that it tells the truth about what it could not read, and "high confidence" on a score where roughly one bar in four is arithmetically impossible undercuts that directly. A musician reading the label will not expect to hit a wrong bar every fourth measure.

Two mitigating facts, which are why this is not urgent:

- the same string appends the actual defective count, so the number is visible next to the label
- the web interface computes the ratio independently rather than trusting the label

So the information is reachable. The problem is that the headline word contradicts it.

Worth deciding rather than just lowering the number: whether a three-way label (high / partial / low) with a much tighter high band communicates better than one boundary, since the current single gate has to serve both "essentially clean" and "mostly usable".

Related: the honesty counters `unknown_at_flag_position` and `undecided_rests` are both **0 across all 2,749 decoded staves** in the library. They are not wrong, but they contribute nothing to the confidence signal on real data, so the label currently rests on bar arithmetic alone — and arithmetic cannot see a chord grouping that is wrong but internally consistent (see #108).

Issue #103, opened 2026-08-22
Provenance we already compute is never shown


sirged:
The server works out whether several values were **read** or **assumed**, and then nothing displays the difference. The distinction exists as data and dies at the API boundary.

Measured across `main` — fields the server reports against whether anything in the interface reads them:

| field | what it records | read by the interface |
| --- | --- | --- |
| `key_signature_source` | whether the key was decoded from the engraving or assumed | **no** |
| `time_signature_source` | whether the meter was read from the digits or assumed 4/4 | **no** |
| `tuning_label` | whether a non-standard tuning was recognised | **no** |
| `local_date_source` | whether a practice day was recorded or inferred from a timestamp | **no** |
| `restored` | how many scores a scan re-found after being marked missing | **no** |
| `sessions_inferred` | how much of an aggregate rests on inferred days | **no** |

Two of those matter a great deal, because 18% of first pages lose their printed meter (#90) and every score not literally labelled "Drop D" is read as standard tuning (#80). In both cases the extractor **already knows** it guessed. The interface shows the guess and not the fact that it is one.

This is the same defect as the one #85 just fixed one layer down: a value that was invented, presented indistinguishably from one that was read. There it was silence inside a bar; here it is the key, the meter and the tuning of the whole score.

## What to do

Surface the read-versus-assumed distinction wherever the value itself is shown. It does not need much — the pattern already exists in the rhythm confidence mark and the warning summary: state the fact quietly, next to the thing it qualifies, without a hover, since a tablet on a stand has no pointer.

For the tuning specifically, #80 already argues the display belongs on the staff at the first bar, where a player looking at their own arrangement will catch a wrong assumption instantly.

## The pattern worth naming

Six fields is not an accident, it is a habit: the server half of a feature gets built and verified, and the half that tells the user gets left. It is how #72 ended up with 35 server references and none in the interface, and how the figures in #101 were computed for consumers that do not exist. Worth watching for at review time, not just here.

PR #118, opened 2026-08-23
Say when a duration was floored, and which bars were not read (#115, #117)

### LINKED ISSUE #115: 1,954 noteheads get a guessed quarter duration and nothing counts them
Found by the adversarial review of #108. Pre-existing — 1,996 at `main`, 1,954 on the branch — so the PR does not introduce it, and the PR states the number honestly.

A filled notehead that ends up with no stem attached has its flags left uncounted and is emitted as a **plain quarter**. Across the library that is **1,954 notes** whose duration is a floor rather than a reading. A flagged or beamed note in that state reads long, and it loses its voice.

The defect is not the guessing — some noteheads genuinely cannot be resolved. The defect is that **nothing counts it**:

- there is no `no_stem` counter anywhere in `server/fermata`
- `decode_note_events`' stats carry `undecided_rests` for a guessed rest and have no equivalent for a guessed notehead duration
- so these 1,954 guesses are invisible to `_resolve_rhythm_source`'s provenance gate and to the warning surface

That means a score can be built substantially on floored durations and still report its rhythm as read from glyphs. This is the same shape as the problems #101 and #103 were filed for: the extraction knows it guessed and does not say so.

Related, and part of why this matters now: #114 records that the confidence label currently rests on bar arithmetic alone, because both existing honesty counters are 0 across the whole library. A counter that actually fires on real data would be the first one.

### Verifying a fix

A counter that reports 1,954 on the current library, feeding the provenance gate and the warning surface the same way `undecided_rests` does. The count must be reachable from `decode_note_events` on real scores — not a branch that only a synthetic input can enter, which #108 already had to clean up once.

sirged:
Closes #115. Addresses #117, with a correction to its premise below.

The server half of the recurring defect: the measurement gets built, and the half that tells the user gets left. Both additions here exist as a structured field **and** as warning prose, because the interface loops over `warnings` as a list of strings and displays them, while a new field is read by nothing written before it existed.

## #115 — nothing counted the guessed note durations

A filled (or x / diamond) notehead that reaches the end of `decode_note_events` with no stem — neither a stem end near it nor a stem passing through it — has no flag or beam to count, because both attach to the stem. It is emitted at the longest value its notehead alone allows: a plain quarter. That is the *floor* of the candidate readings, so such a note always plays long and always pushes its bar towards the overfull side, and it has lost the stem direction that assigns it to a voice as well.

- `decode_note_events` now reports `no_stem_noteheads`.
- A staff carrying any of them resolves to `glyphs-degraded` rather than `glyphs`, with a per-staff reason, fed exactly the way `undecided_rests` is. **Not** ratio-gated, for the same reason `unknown_noteheads` is not: how dense the staff around a note happens to be is not evidence about that note.
- `ExtractionResult.notes_no_stem` / `.staves_no_stem`, persisted into the confidence blob and surfaced on both POST and GET.
- A document-level sentence naming the number of **notes**, which is the only place that number is said — the staff counts beside it cannot be turned back into it, since one stemless notehead on a staff and forty read identically there.

Half and whole noteheads are deliberately excluded: neither shape can carry a flag or beam in any notation, so a missing stem costs the voice signal and nothing about the duration.

**The number does not reproduce.** Measured over the 297 library PDFs (293 extractable):

| definition | count |
|---|---|
| filled/x/diamond heads leaving the decode with no stem — what this PR counts | **2,227** |
| ...of those, reaching a stem group in an emitted bar | 2,132 |
| ...of those, in a group whose flags came out 0, i.e. actually emitted at the floor | 1,805 |
| `notehead_filled` only, stemless after `_assign_stem_directions` | 2,153 |
| filled/x/diamond, stemless after `_assign_stem_directions` | 2,353 |

Nothing lands on 1,954. The last row is above the counter because `_assign_stem_directions` strips `stem_key` from a further ~126 heads *after* their flags were already read — those lose their voice, not their duration, so counting them would inflate the figure.

Distribution: 493 of 2,749 decoded notation staves, in 182 of 293 scores. Among affected staves the ratio of stemless-to-total heads is 7.3% at the median and 20.6% at the 90th percentile, so these are not one-off strays.

**Reachability.** Nothing engraved in this repository reaches this state — MuseScore draws every stem as a clean vector line, and all twelve committed fixtures report zero. The end-to-end test therefore runs against a real library score (`Hymn of the Fayth`: 73 noteheads, 4 staves, all 4 degraded) rather than against a branch no real input can enter, which is the mistake #108 shipped and had removed. That limit is written into `test_engraved_fixtures.py`'s "what these cannot reach" list rather than left implicit.

## #117 — spacing-derived rhythm

**The premise is wrong, and this is the correction.** `rhythm_provenance` over the library is exactly as filed — `glyphs 2656 / glyphs-degraded 3 / spacing 125` — but prose for it already existed and already reached the user. `_rhythm_report` has emitted *"durations were read from the engraved notation for N staff system(s); M staff system(s) could not be read that way and use a rougher estimate from note spacing instead"* since before this branch, and all 5 library scores holding a spacing staff already finish at `low overall`, not `high` - none of them presents as read. `glyphs-degraded` had its own sentence too.

What was actually missing is **where**. "Treat those sections as low confidence" was an instruction a reader had no way to follow: the message counted staff systems and named no music, unlike the padded-bars message it was supposed to match. So:

- `spacing_bars` and `degraded_bars`, collected per staff from the document's own bar numbering, carried on `ExtractionResult`, persisted, and surfaced on GET.
- Both sentences now name their bars through the existing `_bar_list` cap.
- The degraded summary sentence no longer names an uncalibrated glyph as *the* cause — with this PR that is the minority cause — and states the three instead.

One correction to a claim in the issue: `rhythm_provenance` is **not** in the stored confidence blob. It is on the extract response only, so it dies on the first reload. The new lists are stored, so they survive.

**`glyphs-degraded` keeps its own disclosure.** It is not `spacing` — the glyphs were decoded — and it is not `glyphs` — something on the staff was not read. Folding it either way would make one of those two claims false, and it stops being a rounding question after this PR anyway: it goes from 3 staves to 495.

## Effect on the reported label

Measured over 293 extractable library scores, before → after:

```
high    27 -> 19
medium   0 ->  8
low    266 -> 266
```

Eight scores move from `high` to `medium`, all eight for stemless noteheads. Nothing moves out of `low`: 174 of the 182 affected scores were already `low` on bar arithmetic. Provenance goes `glyphs 2656/deg 3/spacing 125` → `glyphs 2164/deg 495/spacing 125`.

## The two inert counters

`undecided_rests` and `unknown_at_flag_position` are both still 0 across all 2,749 decoded staves. Neither is dead code and neither is measuring the wrong thing — they are **genuinely rare, and rare for a stated reason**. `undecided_rests` only fires where a font spells the half and whole rest with one glyph *and* the geometry declines the parity, which needs an unreadable outline or an undetected staff-line grid; `unknown_at_flag_position` needs a glyph outside the calibrated vocabulary sitting exactly at a stem's free end, and the library's unknown-glyph ratio is 0.0 at the 99th percentile. They are tripwires for input this library does not contain, which is a fair thing for them to be — but it does mean that before this PR the confidence label rested on bar arithmetic alone, exactly as #114 records. `no_stem_noteheads` is the first counter that fires on real data.

## Should the new counter feed the confidence gate?

Not in this PR, and #114's thresholds are untouched. It already caps the label through provenance (8 scores `high` → `medium`). Feeding it into the 0.25 defective-bar ratio would be the wrong shape: that gate is a *proportion of bars*, and this is a count of *notes*, so it would need its own threshold rather than joining that one. Worth noting for #114 that the two overlap heavily — a floored duration reads long, which is precisely what overfills a bar, so a large part of what the bar-arithmetic gate already catches is this defect seen downstream. A separate note-level threshold would mostly re-condemn scores that gate has already condemned; the 8 scores that changed label here are the ones it would newly reach, and provenance reaches them already.

## Tests

12 new tests; suite goes 637 → 649 passed, 2 skipped (the two `FERMATA_MUSICXML_XSD` ones, unchanged). Without a library: 622 passed, 29 skipped — 27 library skips, up from 25, the two new ones being the real-score tests above. `npm ci` was run in `web/` so the six alphaTab round-trip tests actually ran.

Every change was proved to be load-bearing by reverting it and confirming red:

| mutation | red |
|---|---|
| counter never increments | 2 failed |
| counter stops excluding half/whole heads | 5 failed |
| provenance gate never fires | 2 failed |
| the count exists but nothing says it | 2 failed |
| spacing sentence stops naming its bars | 1 failed |
| degraded sentence stops naming its bars | 2 failed |
| degraded sentence reverts to naming one cause | 1 failed |
| bar numbers never collected | 3 failed |
| new figures not persisted | 1 failed |

Committed before measuring, since a harness that reverts with `git checkout --` discards uncommitted work and then measures the old code.

## Review findings addressed

Both blockers and every smaller item from review were fixed in place:

- **Blocker 1** (`tabextract.py`, the ≥0.25 defective-bar branch): it REPLACED `confidence` with a fresh "durations were read from the score's own engraving," discarding whatever degraded-stem or spacing disclosure the score had already earned — true on 174 of 182 affected scores. It now composes onto the existing `confidence` (stripping only the old label, the way the sibling branch just below it already did), so the stem/spacing caveat survives into the headline. Proved red against the prior code on the `Hymn of the Fayth` fixture (every staff floored, ratio over the gate) before fixing it.
- **Blocker 2** (the floored-note-durations sentence): it attached an emission-stage claim ("plays long and overfills its bar") to the decode-stage count of 2,227, true of only 1,528 of them. The count still gates on 2,227; the sentence now claims only what holds for every one of them — duration and voice rest on a guess, and where such a head could not be attached to a neighbouring stem it was emitted at the plain quarter.
- The two summary sentences reading "no stem attached to them" / "no stem to count flags on" now read "no stem this decoder could find," matching the per-staff wording — these are Finale/Sibelius engravings; the stems exist, this decoder just didn't find them.
- The uncommitted denominator ("493 of 2,749 notation staves") is now "493 of the 2,657 notation staves that supplied glyph durations at all" — the same population the numerator was measured against — fixed identically in `test_engraved_fixtures.py`.
- PR body corrected: p90 stemless ratio is 20.6% (7.3% is the median, not the p90); 293 extractable scores, not 292; the "any spacing staff already reported mixed" claim replaced with what the review measured — all 5 such scores finish at `low overall`.
- Three cheap tests added: a unit assertion pinning the no-stem gate at the end of `_resolve_rhythm_source`'s ladder (a staff mostly out of vocabulary AND carrying stemless heads must resolve to spacing, not glyphs-degraded); the real-score API round trip extended to assert a non-zero `notes_no_stem` survives `transcribe` → `get_transcription`; `check-warning-patterns.mjs` fixtures for the three score-specific sentences this PR adds or reworded, so none of them can start matching `BAR_RE` and vanish from the per-score list.
- Stale comments fixed: `api.py`'s `_BAR_LIST_KEYS` comment no longer cites `rhythm_provenance` as something this module exposes; `ScoreCompare.svelte`'s save-edit comment now lists all four new keys (`spacing_bars`, `degraded_bars`, `notes_no_stem`, `staves_no_stem`); the Maestro GID table now notes that gid 174 (diamond) is bucketed with `notehead_filled` as `duration_needs_stem`, with quarter-vs-half unpinned for it.

Suite now: 651 passed, 2 skipped (with `FERMATA_TEST_LIBRARY` set and `npm ci` run in `web/`). The three new tests were each proved load-bearing the same way as the rest of this PR:

| mutation | red |
|---|---|
| the no-stem gate moved ahead of the spacing fallback in the ladder | 1 failed |
| the API layer writes 0 into `notes_no_stem` unconditionally | 1 failed |
| a fixture sentence reworded to match `BAR_RE` | `check-warning-patterns.mjs` fails |

Committed before each mutation and reverted with `git checkout --` afterwards, then re-confirmed green.

## Not touched

`web/` — the interface half is #103, in flight separately. It needs no change to pick these up: `ScoreCompare.svelte` renders the whole `warnings` list and filters only the two standing limits, so the new sentences already reach the reader, and `node scripts/check-warning-patterns.mjs` still passes against the reworded degraded message.

### DIFF
--- a/docs/musicxml-tab-profile.md
+++ b/docs/musicxml-tab-profile.md
@@ -841,6 +841,32 @@ into the reported confidence instead. That confidence also states any known
 defect below the threshold at which it changes label, so an unqualified
 high-confidence string means no measure was in question rather than not many.
 
+**How each measure's durations were obtained** is reported alongside them.
+`rhythm_provenance` counts staff systems by source: `glyphs` means flags,
+beams, dots and rest shapes were decoded; `spacing` means durations were
+inferred from the horizontal gaps between noteheads, which is only as good as
+the engraver's spacing being proportional and is wrong wherever a system was
+justified to the margin or spaced by hand; `glyphs-degraded` means the staff
+was read from its engraving with something on it left unread. `spacing_bars`
+and `degraded_bars` are the measure numbers those systems produced. The counts
+say how much of a transcription is in question, and only the lists say which of
+it — which is the form of the fact a reader comparing against the PDF can use.
+
+`notes_no_stem` is how many filled noteheads were read with no stem attached,
+across `staves_no_stem` notation staves. A note's flags and beams hang off its
+stem, so for those notes there was nothing to count and each was emitted at the
+longest duration its notehead alone allows — a quarter. That is a floor rather
+than a reading, and because it is the longest of the candidate values such a
+note always plays long and always pushes its measure towards Rule 8's overfull
+side; it has also lost the stem direction that assigns it to a voice. A staff
+carrying any of them is reported as `glyphs-degraded` rather than `glyphs`.
+
+Every one of these figures is also stated in the warning prose, in the same
+numbers. That is deliberate duplication rather than redundancy: a consumer
+written before a field existed does not read it, and the warnings are a list of
+strings that gets displayed as it is, so a count added without its sentence is
+a measurement nobody is ever told.
+
 **How the meter and key were obtained** is reported the same way, and for the
 same reason: `time_signature` with `time_signature_source`, and `key_fifths`
 with `key_signature_source`. A source of `glyph-decoded` or `auto-detected`
--- a/server/fermata/api.py
+++ b/server/fermata/api.py
@@ -1314,8 +1314,14 @@ def _transcription_row(conn, score_id: int):
 # holding silence that was deduced from the meter rather than read from the
 # page, or holding nothing that was read at all, is not the reading the other
 # figures make it look like, and neither is recoverable from them.
+#
+# `notes_no_stem` / `staves_no_stem` belong with them too. A filled notehead
+# whose stem the decoder never found has no flag or beam to count, so it was
+# emitted at the longest duration its notehead alone allows - a floor, not a
+# reading, and one that always errs long. That is the same kind of fact as a
+# padded bar and is recoverable from nothing else here.
 _BAR_KEYS = ("bars_overfull", "bars_short", "bars_defective", "bars_measured",
-             "bars_padded", "bars_unread")
+             "bars_padded", "bars_unread", "notes_no_stem", "staves_no_stem")
 
 # WHICH bars those were, as data and not only inside the warning prose. The
 # prose names them, but it caps the list, and the profile document states that a
@@ -1323,7 +1329,13 @@ _BAR_KEYS = ("bars_overfull", "bars_short", "bars_defective", "bars_measured",
 # `inferred_rest_quarters` - a claim the application has to actually make good
 # on. `padded_bars` / `unread_bars` are lists of bar numbers;
 # `inferred_rest_quarters` is a quarter-note count and can be fractional.
-_BAR_LIST_KEYS = ("padded_bars", "unread_bars")
+#
+# `spacing_bars` / `degraded_bars` are the same kind of list for the same
+# reason: which bars' durations came from the gaps between noteheads rather
+# than the noteheads, and which came off a staff something on it could not be
+# read. The staff counts in `rhythm_provenance` say how much; only these say
+# which, and a bar number is what a reader can carry back to the PDF.
+_BAR_LIST_KEYS = ("padded_bars", "unread_bars", "spacing_bars", "degraded_bars")
 _BAR_AMOUNT_KEYS = ("inferred_rest_quarters",)
 
 
@@ -1515,6 +1527,10 @@ def transcribe(score_id: RowId, body: TranscribeIn | None = Body(default=None)):
             "padded_bars": result.padded_bars,
             "unread_bars": result.unread_bars,
             "inferred_rest_quarters": result.inferred_rest_quarters,
+            "notes_no_stem": result.notes_no_stem,
+            "staves_no_stem": result.staves_no_stem,
+            "spacing_bars": result.spacing_bars,
+            "degraded_bars": result.degraded_bars,
             "time_signature": list(result.time_signature) if result.time_signature else None,
             "time_signature_source": result.time_signature_source,
             "key_fifths": result.key_fifths,
--- a/server/fermata/api.py
+++ b/server/fermata/api.py
@@ -1333,8 +1333,10 @@ _BAR_KEYS = ("bars_overfull", "bars_short", "bars_defective", "bars_measured",
 # `spacing_bars` / `degraded_bars` are the same kind of list for the same
 # reason: which bars' durations came from the gaps between noteheads rather
 # than the noteheads, and which came off a staff something on it could not be
-# read. The staff counts in `rhythm_provenance` say how much; only these say
-# which, and a bar number is what a reader can carry back to the PDF.
+# read. `rhythm_provenance` on the extraction result says how many staves
+# resolved each way, but that field is not one this module stores or exposes
+# - only these bar-number lists carry the fact out to a consumer, and a bar
+# number is what a reader can carry back to the PDF.
 _BAR_LIST_KEYS = ("padded_bars", "unread_bars", "spacing_bars", "degraded_bars")
 _BAR_AMOUNT_KEYS = ("inferred_rest_quarters",)
 
--- a/server/fermata/glyph_rhythm.py
+++ b/server/fermata/glyph_rhythm.py
@@ -178,6 +178,13 @@ MAESTRO_GID_MAP = {
     # table's own "rendered and eyeballed" standard. A signature that would
     # need digit0 correctly falls through to "not detected" (see
     # decode_time_signature) rather than silently emitting a wrong value.
+    #
+    # gid 174 (notehead_diamond, harmonics) is bucketed with notehead_filled
+    # as of #115: `duration_needs_stem` is True for it, so an unfound stem
+    # floors it at a quarter the same way a filled head does. Quarter-vs-half
+    # is unpinned for a diamond - unlike notehead_half it carries no shape
+    # that says which, so a diamond engraved as a half is read as a quarter
+    # whenever its stem is not found, not just left undecided.
 }
 
 # Fingerprint for the calibrated Maestro subset: GID -> sha256 of that
--- a/server/fermata/glyph_rhythm.py
+++ b/server/fermata/glyph_rhythm.py
@@ -2087,7 +2087,10 @@ def decode_note_events(page, staff_top, staff_bottom, staff_x0, staff_x1, line_y
     where a flag would attach - the shape of "this piece uses 32nd flags,
     grace notes or an articulation we never calibrated", which decodes as
     systematically wrong durations while looking perfectly healthy - and
-    undecided_rests counts rests whose value was guessed rather than read.
+    undecided_rests counts rests whose value was guessed rather than read,
+    and no_stem_noteheads counts filled noteheads that came out of the decode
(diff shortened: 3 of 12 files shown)


sirged:
## Review findings — two blockers, both in the prose layer; the mechanics survived attack

The counter, the gate, the half/whole exclusion and the persistence were attacked and held: the gate's blast radius is proportionate (median flagged staff has 3 floored of 41 decoded events; the "one stray note degrades 200 good ones" case essentially does not occur in this library), the #70 spread bug is genuinely closed (hand-edited rows state the new keys as unrecorded rather than omitting them), and a chord member whose duration was recovered from a sibling's stem is correctly not counted. The label effect reproduces exactly: high 27→19, medium 0→8, low unchanged.

### Blocker 1 — the headline confidence still affirms the exact claim this PR withdraws

`tabextract.py:2128-2132`: when the defective-bar ratio crosses 0.25, the branch **replaces** `confidence` with a string beginning "low overall - individual durations were read from the score's own engraving, but …" — discarding the degraded-stem sentence this PR just wrote. Measured: **174 of the 182 scores with floored durations end up with a headline that says their durations were read.** Only 8 mention a stem at all.

The PR's own end-to-end fixture demonstrates it: every staff on that file is floored, `notes_no_stem 73`, and the headline the interface shows says the durations were read from the engraving.

The wording predates this PR — but before it, only 3 staves in the library were degraded, so it was near-enough true. This PR is what makes it false 174 times. The sibling branch at 2133-2141 already composes instead of replacing (`f"{confidence}; {reason}"`); do the same here. This is the recurring shape: a guard whose fallback restores the thing being guarded against.

### Blocker 2 — the prose attaches an emission-stage claim to a decode-stage count

The warning says each of the counted heads "was emitted as a plain quarter … plays long and overfills its bar". Traced head-by-head through `_stem_groups` and `_stem_group_duration`, of 2,227 counted heads: **1,528 (69%)** actually emitted at the plain quarter; **265** emitted at an eighth or *shorter*; **247** emitted *longer* (dots and bases inherited from a host group); **187 never emitted at all**. The mechanism is `_stem_groups`' absorb pass: a stemless head is a singleton with no signal, skips the signal-mismatch guard, and is folded into any host within onset tolerance — the host's beam then supplies its duration.

101 of 182 affected scores have at least one head the sentence misdescribes, and several are badly off ("Where the Egg of Dreams Hatches" says 21, zero at the floor). The bar-arithmetic consequence also overstates ~2.8×: the emitted heads occupy 1,004 distinct beats, 796 at the quarter.

2,227 is the honest number for the **gate** — the stem was not found, so both duration and voice rest on a proximity guess either way. It is not the honest number for a sentence about what was **emitted**. Soften the consequence clause, or carry a second emission-stage count for the prose. Note "1,805 emitted at the floor" from the PR body is measured on `flags` alone; on the emitted `(code, dots)` it is 1,528.

### Smaller, all wanted before merge

- Two summary sentences blame the engraving ("noteheads … with no stem attached to them"). These are Finale/Sibelius engravings; the stems exist. The per-staff detail already gets it right ("no stem this decoder could find") — align the other two.
- Committed figure that does not reproduce: "493 of 2,749 notation staves" — the 493 is right, 2,749 matches nothing measured (2,657 notation staves supplied glyph durations); the comment also mixes a notation numerator with what looks like a tab denominator. Same wrong denominator in `test_engraved_fixtures.py:38-40`.
- PR body corrections: p90 stemless ratio is 20.6% (7.3% is the median — the understatement direction, but wrong); 293 extractable, not 292; the "any spacing staff already reported mixed" claim — all 5 such scores actually finish at "low overall".
- Test gaps, each cheap: (a) the gate's position at the end of the resolution ladder is deliberate and unpinned — moving it ahead of the spacing fallback leaves the suite green; wants a unit assertion, not a code change. (b) No test asserts a **non-zero** `notes_no_stem` surviving the API round trip — a persistence bug writing 0 unconditionally passes everything; run the real-score test through the API. (c) The new score-specific sentences have no `check-warning-patterns` fixtures, which is the guard against a reworded sentence matching `BAR_RE` and vanishing from the per-score list.
- Stale comments: `api.py:1336` cites `rhythm_provenance` as if the API exposed it; `ScoreCompare.svelte:395`'s key list not extended. One line wanted in the Maestro calibration comment for gid 174 (diamond, now `duration_needs_stem`, quarter-vs-half unpinned).

Issue #121, opened 2026-08-23
6/8 metronome accents do not match what a player expects under subdivision


sirged:
From human testing of the general metronome (which otherwise "works great"): in 6/8, the accent pattern sounds wrong under subdivision — "it seems to accent on the 6 still instead of for 8ths", tried with both 8th and 16th subdivision settings. Confidence from the tester was "not 100%", so the first job is to establish ground truth, not to patch.

## What the code says should happen

`metronomePattern(6, 8)` (`web/src/lib/metronome.js`) returns `clicksPerBar: 6, accentEvery: 3` — compound-meter handling, accents on clicks 1 and 4, the two dotted-quarter beats. The engine multiplies both by the subdivision (`metronome-engine.js:175-176`), which preserves those accent *positions* at any subdivision depth. If the engine actually receives (6, 8), the accents land correctly by construction.

So the report implies one of:

1. **The page's meter controls do not deliver (6, 8) to the engine.** If the standalone page normalises the meter, or its denominator control does not reach `metronomePattern`, `isCompound` is false and `accentEvery` becomes the full bar — one accent per bar, which a player would describe exactly as "accents on the 6".
2. **The subdivision labels mislead in compound meter.** In 6/8 the base click already is the eighth. If the subdivision picker's labels assume a quarter-note beat ("8ths" = split in two), then in 6/8 "8ths" either does nothing or means something a player does not expect.
3. **The pattern is right and expectation differs** — some players want the subdivided clicks between accents shaped (beat-level tick vs subdivision-level tick as three distinguishable levels, not two).

## What is wanted

First reproduce with instrumentation, not ears: drive the engine at 6/8 with subdivision 1 and 2 through the *page's own controls* and log which phases get `accent: true`. That settles which of the three cases this is. Then fix accordingly — and if it is case 3, that is a design decision to surface, not silently pick.

Coverage note: any fix needs a test asserting accent *positions* for a compound meter through the page's control path, not just `metronomePattern`'s return value — the unit is demonstrably right while the sounding behaviour is reported wrong, so the gap is between them.

sirged:
## Reproduced, instrumented. The accent *positions* are right; the accent *sound* and the subdivision *labels* are both wrong. Hypothesis 1 is out, 2 and 3 are both true and together they produce exactly what was reported.

[...]

Issue #104, opened 2026-08-22
The meter is resolved per staff system, so a mid-system change budgets earlier bars against the wrong one


sirged:
The meter a bar is measured against is resolved **per staff system**, not per
bar. In `tabextract._extract`:

```python
anchor_y = std_staff.top if std_staff is not None else staff.top
staff_ts = _ts_at(ts_timeline, page_idx, anchor_y) or ts
measure_quarter_len = _measure_quarter_length(staff_ts)
```

`anchor_y` is the notation staff's own top, so one meter is chosen for the whole
system and every bar in it is budgeted against that. A time signature printed
part-way along a system therefore applies to bars *before* it as well as after,
because `_ts_at` is asked about the system's position rather than the bar's.

Two things follow, both of which land on the figures this project reports:

- The wrong meter gives the wrong budget, so `_bar_conformance` measures those
  bars against a length the engraver never wrote — they can be reported
  overfull, short, or conformant for no reason connected to how well they were
  read.
- `_pad_voice_to_budget` fills a short voice up to that same budget, so a bar
  ahead of a mid-system change can be padded with silence towards a meter it is
  not in.

**Scale.** 16 scores in the library change meter part-way through. They carry
very high defect rates, which is consistent with this but does not demonstrate
it: those scores are also the ones with the most going on. No specific bar has
been shown to be padded wrongly for this reason, so treat the scale as unknown
until one is.

**What is not wrong.** This is not a reporting disagreement. The figures
reported still equal, exactly, what an independent Rule 8 check computes from
the emitted MusicXML — the file is measured against the same (possibly wrong)
`<time>` it declares. So a consumer and this extractor agree; they may simply
agree about the wrong meter.

**What a fix looks like.** Resolve the meter at the bar's own x/y position
rather than the system's — `_ts_at` already takes a position, and
`_build_time_signature_timeline` already records where each signature was read.
The measure loop knows `m_lo`/`m_hi`, so the information is present; it is the
lookup that is too coarse.

Found while working on #85 (a review of that PR raised it). Deliberately not
fixed there: it changes which meter bars are measured against, which would move
the conformance figures for reasons unrelated to inferred silence and make that
change impossible to read.

sirged:
[...]

PR #122, opened 2026-08-23
Read the meter that is printed, and bar each bar in it (#90, #104)

### LINKED ISSUE #90: The time-signature search stops before the key signature, and a later meter is then applied retroactively
Two defects that compound into one wrong answer, stated at high confidence.

## 1. The search window is too short to reach the meter

The time-signature digits are looked for within a fixed distance of the staff's left edge — a window that has to accommodate the clef, the key signature *and* the meter. Three or more accidentals push the digits outside it, and the printed meter is simply not found.

Measured: **49 of 275 tab-bearing first pages — 18% — lose their printed time signature this way.** The discriminator is crisp rather than marginal: every staff with three or four accidentals in the window fails, every staff with two or fewer succeeds.

Concretely, a library score has its clef at 1.1 spaces, four sharps from 4.9 to 8.0, and the 4/4 digits at 10.5 — beyond the window. The decoder reports finding zero digits and the extractor falls back to assuming 4/4.

**This is the direct cause of the many "not detected (assumed 4/4)" results.** And the fallback is not harmless where the real meter differs: 6/8 pieces (`Eternity — Memory of Lightwaves`, `Score A`, `Score B`) and 3/4 pieces (`Score C`, `Score D`) are each barred as 4/4 for their entire transcription, which misplaces every barline and makes every conformance figure meaningless.

The window should extend past the key signature — keyed off the last accidental's position rather than a flat distance from the staff edge.

## 2. When the opening system fails, a later meter is backdated over it

The opening meter is taken as the **first successfully decoded** one anywhere in the score. Staves whose decode returns nothing are skipped silently.

Given defect 1, an opening system failing is routine. So for a piece in 4/4 that changes to 3/4 partway through, the only decoded entry is the 3/4 one — and that becomes the opening meter. Every earlier bar is then measured and emitted in 3/4, at confidence "high — read directly from the time-signature digit glyphs printed on the score", against music that was written in 4/4.

Worse, because only one meter was ever recorded, the "this score changes time signature part-way through" warning **never fires**. So the one signal that would prompt someone to look is suppressed by the same failure that caused the problem.

The comment directly above that code states the design intent as avoiding "taking whichever staff answered first" — which is precisely what indexing the first entry does.

## Why together

Either alone is a bounded problem. Together they turn a common engraving choice — a key signature with three sharps — into a silent, confident, whole-score meter error with its warning suppressed. Fixing the window makes the second defect much rarer; fixing the backdating makes it honest when it happens. Both are wanted.

A useful invariant for whoever fixes this: a decoded meter that applies to bars *earlier* than the staff it was read from should never be reported at full confidence, and a score where some staves decoded and others did not should say so.

Related: #84 (unmapped digits producing a confidently wrong meter — same output, different cause), #88 (the numerator/denominator split relies on a biased vertical measurement).

sirged:
Fixes #90 (both halves) and #104. Three defects that compound: the first
makes the second routine, and the third corrupts the budget even once both are
fixed. Also fixes a blocker an adversarial review found in this branch's own
mid-system reader (section 4), and corrects two claims the same review showed
were stated as measured when they were not (sections 1 and 3).

## 1. The search window could not reach the meter (#90)

The digits were looked for within a flat 8.8 staff spaces of the staff's left
edge — a window that has to hold the clef, the key signature *and* the meter.
Confirmed on the issue's own example, `Score R`: clef at 1.08 spaces, four sharps from 4.87 to 8.00, the 4/4 digits at 10.54.

The window now ends at whichever comes first of a fixed reach past the last
key-signature accidental and the first glyph on the staff that **sounds**.

**Measured, before → after.** First pages carrying both tab and notation:
292. Losing their printed meter: **53 → 4**. All 49 that failed on the window
are read now; the 4 that remain report "no music glyphs found" — no readable
music font on the page at all, which no window reaches. 49 scores went from
"assumed 4/4" to decoded, and **12 of them were not in 4/4**.

The named scores, each checked against the page image and not only against
the decoder: `Score A` 6/8, `Score B` 6/8, `Eternity — Memory of Lightwaves` 6/8, `Score Q` 3/4 (three sharps), `Score C` 3/4 (five sharps). `Score R` is genuinely 4/4 — the
window is why it was not *read*, and it now is.

**One figure from the issue does not reproduce.** The denominator: I count 292
first pages with both a tab and a notation staff, not 275, and 53 losing the
meter of which 49 are window failures — so the 49 is exact and the ratio is
49/292. The discriminator is also not quite crisp: on `main`, staves with 3
accidentals in the window lost the meter 28 times and found it **twice**, 4
accidentals lost 18 and found once, and there are staves with 5 and 6
accidentals as well (3 lost, 1 found). Tighter engraving at a wider staff
spacing is enough to get a four-sharp meter inside 8.8 spaces. My accidental
count is taken over a 20-space window, so it can include an accidental on the
first note; that inflates the high end rather than the discriminator.

**Corrected after an adversarial review.** The paragraph here used to claim
the notehead-or-rest bound "is free" and "closes" the hazard of a widened
window reading a fingering or tuplet number as a meter, as established fact.
It is not measured to do either: removing the clamp, flattening the window to
a plain 20 spaces, or both, changes **0 of the library's 297 scores**. That is
not the same as the clamp being useless — see section 4, where the identical
construction on the mid-system reader IS measurably load-bearing (Kaine
Salvation) — only that this specific library never exercises the difference
on the OPENING window. Kept for the reason section 4 gives, not because this
library proves it necessary here.

The clamp was also blind in one direction nothing above mentioned: it read
"does anything sound" from the same ±1 staff-space band as the staff itself,
which never sees a note on a ledger line — guitar's open low strings (E, A, D
on a treble-8vb staff) sit at or just past that boundary. Its own y-band is
now wider than the staff (`_SOUNDING_BAND_SPACINGS`, measured against the
library's deepest ledger-line noteheads), and two new unit tests cover both
directions: a stray digit pair beyond such a note is refused, and a real
meter printed before one is still read.

## 2. A later meter was backdated over the opening (#90)

The opening meter was the first successfully decoded one *anywhere*, and a
staff that decoded to nothing was skipped silently — so a piece in 4/4 that
changes to 3/4 recorded only the 3/4, reported it as the opening meter at
"read directly from the time-signature digit glyphs" confidence, and measured
every earlier bar against it. And having recorded one meter, it never fired
the "changes time signature part-way through" warning: the one signal that
would prompt somebody to look was suppressed by the same failure that caused
the problem.

The opening meter is now the one printed at the score's **own first notation
staff** and only that one — the stated intent of the comment above that code.
Where it cannot be read, the score says so, keeps the assumption it is barred
with, and reports it as an assumption; the meter that *was* read applies from
where it was read. The change warning is derived from the meters the
transcription is actually barred in — assumed opening included — so it fires
on exactly the case that used to silence it.

The warning is now quiet, though, when the later meter that was read happens
to equal the assumed default: "the opening was not read, but a 4/4 was found,
so bars are barred as 4/4" is not a discrepancy worth a reader's attention. It
also no longer interpolates the internal `time_signature_source` token into
the sentence — that token can itself be the string `"not detected (assumed
4/4)"`, which nested inside prose used to read as "barred as 4/4 (not
detected (assumed 4/4))", a token dump rather than an explanation. A new
engraved fixture (`hidden_opening_meter_matches_the_default`) covers the
suppression directly.

Separately, and left as a documented hazard rather than fixed: `first_staff`
in `_build_time_signature_timeline` names the first notation staff among
pages that already carry a tab staff, not the score's first notation staff in
general. A score whose real opening page is notation-only — a foreword before
the tabbed arrangement begins — is invisible to this timeline entirely, so a
meter printed only there is never read as the opening one; it is not read at
all. Currently unreachable (no score in this library has that shape), but
named in `_build_time_signature_timeline`'s docstring for the day one does.

**Measured:** scores warning that the meter changes **16 → 43** (see the
aggregates table below for why this and the confidence figures differ from
what an earlier version of this PR body reported — main moved under the PR
between when those numbers were taken and now). That rise is not the warning
getting louder: most of the newly-warning scores print a change part-way
along a system that nothing could read before, and the rest print a change on
a system whose meter used to fail the window.

Confidence never rises on a score whose meter is still assumed.

## 3. The meter was resolved per system, not per bar (#104)

`_ts_at` was asked about the staff system's position, so one meter was chosen
for a whole system and every bar in it budgeted against that. Timeline entries
now carry the x they were read at, and the lookup takes the bar's own left
edge.

**A bar demonstrated wrong before the fix**, as the issue asks. On **Score D** the second system carries a 4/4
engraved part-way along it, and bars 4 and 5 come *before* it and are printed
in 3/4 (verified against the page). Resolved per system they are budgeted 4.0
quarters, read short, and padded towards a meter they are not in; per bar they
are budgeted 3.0. Score-wide, that one difference moves short bars 13 → 11,
padded bars 12 → 11 and inferred silence 9.0 → 7.0 quarter notes.

Making that reachable at all needed the meter *change* to be readable
mid-system, since a change engraved there is printed once and never reprinted
at the next system's start. **41 scores in the library print one**, and five
were checked against the page: `Score E` (a 6/4 bar at the fourth bar of a system), `A Pendant Darkly` (5/4 then back), `Score F` (12/8), `Score D` (4/4) and `Score G` (which alternates constantly). All five are real.

That reading is deliberately narrow: only just past a barline, only a
numerator run stacked over a denominator run, and only when nothing that
sounds comes before it in the window. Most of what a notation staff offers as
a barline is a stem — **21,680 candidate positions** across the library, and
(after the section 4 fix) 152 meters accepted, 10 refused for sitting behind
a note or a rest.

**Corrected after an adversarial review — the "10 refusals" story overstated
itself.** Removing the guard changes nothing on any score in the library: for
all 10 refusals the refused value equals the value accepted at the real
barline **1.7–3.5 staff spaces to the right** — a couple of noteheads away,
not, as this section used to say, "a later barline" a bar away. Nothing is
lost either way because `_append_ts` collapses a repeated value; the guard is
insurance against a hypothetical case where the two values *disagree*, not a
fix for an observed one. The covering test now asserts the refused value
itself (computed by bypassing the guard on the same window), that the NEAREST
accepted candidate to its right — not merely a later one — reads a meter, and
that the two agree.

The docstring's motivating example for the guard was also wrong: two triplet
"3"s stacking into 3/3 cannot reach this guard at all, since 3 is not a valid
denominator and `time_signature_is_valid` rejects it upstream. The real
hazard is an UNEQUAL pair — a triplet "3" in one voice against a "4" or "8"
from a different tuplet bracket, fingering, or string number in another —
forming a numerically valid-looking meter that is not one. Rewritten.

The guard is also blind on the up-stem side, which the review measured rather
than guessed at: an up-stem attaches at its own notehead's RIGHT edge, so the
notehead that owns a stem mistaken for a barline sits to its LEFT — outside
the window this guard ever looks at. Measured across the library: 23 of the
152 accepted mid-system meters have a notehead within one staff spacing to
the left of the candidate. That is not 23 hazards, though — it is what a
GENUINE barline looks like too, since the previous bar's last note routinely
ends that close to it. Left undocumented before, and left unfixed now:
proximity on the left does not distinguish "this candidate is that note's own
stem" from "this is a real barline with an ordinary note before it", so a
look-back guard built on it would refuse correct meters about as often as it
caught wrong ones.

## 4. The mid-system reader could drop a meter behind a key change, or read a courtesy one early

Found by the same adversarial review, in this branch's own new code: the
mid-system reader in section 3 used a flat 5.0-space reach past the barline,
with no accidental-anchored extension the way the OPENING window (section 1)
has. A key change at the same barline as a meter change pushes the meter's
digits out exactly the way it does at a staff's own start — the same defect
section 1 fixes, reintroduced in the reader section 3 adds. Live in the library: `Score H` page 2, a double barline behind
three flats, prints 6/4 at **6.18 staff spaces** past the barline, past the
flat 5.0-space reach. The branch emitted those bars as 2/4 — silently and
permanently, since a mid-system meter is printed once and never reprinted.
Engraved fresh (`mid_system_key_and_meter_change`), four sharps push a 3/4
change out to **5.77 spaces** — the same key-signature size as this PR's own
`four_sharps_in_three_four` fixture, and also past the flat reach.

The 152 meters currently accepted mid-system sit up to 4.92 spaces against
the 5.00-space limit — no headroom — and nothing pinned the constant: moving
it from 5.0 to 9.0 left the whole suite green.

The fix shares the opening window's construction: reach past the last
accidental printed just after the barline, capped, and clamped by the first
thing that sounds (the same widened clamp band from section 1). Widening it
is not safe on its own, though — this is the shape that forbids doing so
naively. `Score I` page 2 prints a *courtesy* 6/8 behind four
sharps about 7 staff spaces past a system's LAST barline, previewing the
meter the next system opens in. That sits well within the widened reach of an
earlier barline in the same system, and the old end-of-system guard keyed
only on the triggering BARLINE's distance from the staff's right edge, never
on the decoded METER's own position — so widening the reach without also
fixing that would have started Kaine's 6/8 a whole system early.

The guard now keys on the meter's own decoded position instead: refused when
it lands within a fixed margin of the staff's right edge. Measured: Kaine's
courtesy signature ends 0.83 staff spaces short of the staff's right edge; Score H's real mid-system 6/4 ends 55.8 spaces short of it — two orders of
magnitude apart, so the margin sits generously between them. The real 6/8 in
Kaine is still read correctly, at the next system's own opening.

Both scores are covered by library-gated tests asserting the exact emitted meter sequence (Score H: `4/4×25, 2/4×1, 6/4×2, 4/4×23`; Score I: `3/4` from bar 1, `6/8` from bar 28, with the courtesy signature checked to be a real,
refused candidate rather than an untested one) and by the new engraved
fixture for the key-change shape. The window constant and the end-of-system
guard are each proved load-bearing in the Tests section below.

## Aggregates on the library

Re-measured after rebasing onto current `main`, which has moved twice since
this PR was cut (#118's honesty counters, #121's confidence-label rework) —
the before-figures below are `main`'s CURRENT numbers, not the ones an
earlier version of this PR body quoted, and they differ from that version in
the confidence row specifically (#121 alone drops 8 scores from high to
medium regardless of this PR).

| | before (`main`) | after |
| --- | --- | --- |
| bars | 10,946 | 10,946 |
| defective | 5,863 | **5,477** |
| overfull | 1,896 | **1,754** |
| short | 4,435 | **4,193** |
| sounding notes | 98,688 | 98,688 |
| bars read from nothing | 337 | 337 |
| scores at high rhythm confidence | 19 | 23 |
| scores warning of a meter change | 16 | 43 |

**Direction.** Defective fell rather than rose. Barring in the right meter
changes what conformance *means*, so either direction would have been
possible; what makes this a real improvement rather than a bookkeeping
shuffle is where it lands. Of the 293 extractable scores: **249 are untouched
to the bar, 38 are better, 1 is worse, and 5 are net-neutral** (same
`bars_defective`, but the overfull/short split or the inferred-silence total
moved under it) — corrected from an earlier "241 untouched" that did not
reproduce under any of nine definitions tried.

The one worse score: a bar of `A Pendant Darkly` that is genuinely in 5/4 and
is now measured as such (10 → 11 defective bars). The score with the largest move, `Score J`, goes from 42 of 44 bars defective to 8, because it is
in 3/4 and was being barred as 4/4. The note count does not move at all,
because none of this reads a note differently — only which meter its bar is
measured against.

The 5 net-neutral scores, named: `Score E` (14 defective either way; overfull 5→4, short 9→10, inferred silence 11.0→13.0 q), `Score K` (28 either way; overfull 7→6, silence 29.75→26.75 q), `Score L` (36 either way; overfull 16→17, short 25→24), `Score M` (6 either way; overfull 1→2, short 5→4, silence 4.75→3.0 q), and `Score N` (9 either way; overfull 4→3, short 5→6, silence 1.0→5.0 q).

The 19 → 23 rise in high rhythm confidence is the conformance ratio, not a
new claim: a score whose bars now add up is no longer downgraded for not
adding up. None of the 23 is a score whose meter is assumed.

**The aggregate hides where part of the improvement was bought.** Inferred
silence — quarter notes padded in because a voice fell short of its bar's
budget, not read from a printed rest — ROSE on 15 scores by 119.0 quarters
combined, even though the library-wide total fell (5,529.25 → 4,835.75 q).
It concentrates in the headline wins: `Score O` goes from 39.75 to 81.0 q of inferred silence, because correcting its bars'
budgets to the meters actually printed makes voices that were previously
measured against the wrong (shorter) length now fall visibly short of the
right (longer) one. Mechanically honest — the silence is marked and excluded
from the Rule 8 conformance count either way — but it is part of the price of
the defective-count improvement above, not a free side effect of it.

## Tests

Four engraved fixtures, because no committed fixture had any of these shapes:

- `four_sharps_in_three_four` — four sharps between the clef and the meter,
  and in 3/4 so that failing to read it misplaces every barline rather than
  landing on the right answer by luck.
- `hidden_opening_meter` — an invisible opening `<time>` and a printed 3/4
  later, so the only meter that can be read is not the opening one. This is a
  real editorial choice, and the one shape that reproduces "one staff decodes
  and the first one does not" without depending on the window defect.
- `mid_system_meter_change` — a change to 2/4 part-way along the first system
  and back to 4/4 at the second, every bar adding up exactly to its own meter,
  so a bar budgeted against its system's meter cannot come out conformant.
- `mid_system_key_and_meter_change` — added after the adversarial review: a
  key change at the SAME mid-system barline as a meter change, the shape that
  reproduces section 4's blocker (5.77 spaces past the barline) without a
  real-library file.

One more, added for the same reason:

- `hidden_opening_meter_matches_the_default` — the opening meter is invisible
  and the only meter ever read anywhere is a 4/4 that happens to equal the
  assumed default, so the "opening not read" warning has nothing to warn
  about and must stay quiet (section 2).

Each fixture's per-measure meters are asserted against the MusicXML it was
engraved from, and the mid-system ones assert the change really is inside the
first system, so a re-engraving that moved the system break could not quietly
turn that test into a tautology. A parametrized test in the other direction
holds six single-meter fixtures — including the ones printing a triplet
number and "1." / "2." endings — to one meter throughout.

Two library-gated tests cover section 4 directly against the real scores:
one asserts Score H's exact emitted meter sequence, the other asserts Score I's courtesy signature is refused as a real, checked candidate (not an
untested absence) while the real 6/8 is still read at the next system's own
opening.

Two synthetic unit tests cover the notehead/rest clamp's widened band in
each direction (section 1): a stray digit pair beyond a ledger-line note is
refused, and a meter printed before one is still read.

Two more cover `_ts_at` directly: one feeds it an unsorted timeline and
checks the answer is unchanged from the sorted case; one pins the documented
`anchor_y` hazard (an anchor even a hair past a system's own recorded y
silently returns that system's LAST meter for a query about its first).

**Proved able to fail.** Each fix reverted separately, whole suite, library
and `npm ci` in place:

| mutation | red |
| --- | --- |
| flat window back in `decode_time_signature` | 1 — `test_a_meter_behind_a_key_signature_is_still_read` |
| opening meter back to whichever staff answered first | 1 — `test_a_meter_read_later_is_not_backdated_over_an_unread_opening` |
| meter resolved per system instead of per bar | 2 — the mid-system test, and the hidden-opening one (its courtesy 3/4 spreads back over the whole first system) |
| notehead-or-rest guard removed from the mid-system reader | 1 — `test_a_meter_further_along_a_bar_is_not_a_meter_at_this_barline` |
| mid-system window's accidental-anchored reach removed (flat 5.0 restored) | 2 — the Score H test and the `mid_system_key_and_meter_change` fixture test |
| mid-system end-of-system guard removed | 1 — the Kaine courtesy test |
| notehead/rest clamp's widened band narrowed back to ±1 space | 1 — the ledger-line-note clamp test |

Suite: **649 passed, 8 skipped on `main`** (measured directly, not carried
over from an earlier version of this body) **→ 689 passed, 2 skipped** with
the library set and `alphaTab.mjs` built (the 2 want `FERMATA_MUSICXML_XSD`).

**Timing, softened.** An earlier version of this PR body reported specific
before/after millisecond figures for `Score G` and `Score P`.
Those did not reproduce on this machine at anything resembling the claimed
precision — the direction is not resolved above measurement noise here — so
the specific numbers are withdrawn rather than restated. Nothing in this PR's
changes does asymptotically more work per staff than before; a proper
before/after timing claim would need a machine and methodology dedicated to
that measurement, which this review pass was not.

## One doc figure

`docs/musicxml-tab-profile.md` quotes 5,863 defective measures inside a
MuseScore round-trip measurement. The producer-side half of that pair is now
5,477 (unchanged by section 4's fix — re-checked after rebasing and after the
blocker fix); the round trip has not been re-measured, so the paragraph is
annotated rather than half-updated with a number nobody took.


sirged:
## Adversarial review — one blocker, two unmeasured claims; the core fixes held

**What held under attack:** the per-bar lookup is real and load-bearing (reverting it moves 38 scores and reddens 2 tests), the x semantics are sound in the safe direction (a false read inside a bar takes effect only from the next barline), C/cut-C and 12/8 mid-system read correctly, the six-fixture no-invented-change test constrains the actual per-measure MusicXML rather than the warning string, and padding moves coherently with the new budgets. Nearly every figure reproduced exactly — including the accidental table, 53→4, 16→44 warnings, and the 1-worse score.

### Blocker — the mid-system reader reintroduces this PR's own headline defect

The new mid-system window is a flat 5.0 spaces past the barline. A key change at the same barline as a meter change puts accidentals between the two — exactly the geometry #90 was about — and the digits fall outside the window. The change is then dropped **silently and permanently**, since a mid-system meter is printed once and never reprinted.

Live in the library: `Into the Wilderness (Wild Arms)` page 2 — double barline, three flats, printed 6/4 at 6.18 spaces. The branch emits those bars as **2/4**, a three-fold budget error, and the change warning names the wrong meters. Engraved fresh: 3 flats reads, **4 sharps misses** (5.77 spaces) — the same key-signature size as this PR's own `four_sharps_in_three_four` fixture. And the 225 accepted mid-system meters sit up to 4.92 spaces against the 5.00 limit: no headroom, and nothing pins the constant (5.0→9.0 leaves the suite green).

The opening window already solves this — reach past the last accidental, capped, clamped by the first sounding glyph — and the mid-system reader should share it. **But do not simply widen it:** `Kaine Salvation (NieR)` p2 has a *courtesy* 6/8 behind four sharps at ~7 spaces past the system's **last** barline, and the end-of-system guard keys only on the barline's distance from the staff edge, never on the meter's own position. Widening without an end-of-system rule starts that change a bar early. Both scores are the test cases.

### Two claims stated as measured that measurement contradicts

- **The notehead/rest clamp and the accidental-keyed reach change nothing.** Dropping the clamp, flattening the window to 20 spaces, or both: **0 of 297 scores differ**, suite stays green. Defect 1's fix, on this library, is "8.8 → 20 spaces". The docstring's "the second half is what makes widening safe rather than merely wider" asserts what no test or measurement supports. The clamp may still be right — F3 below says it is currently *blind* — but it needs a regression test in each direction, not prose.
- **The "10 refusals" story does not survive.** Removing the guard changes nothing on any score; for all 10 the refused value equals the accepted meter 1.7–3.5 spaces to the right — a couple of noteheads away, not "a bar early" — and the duplicate collapses. The covering test asserts only that *some* later meter exists. Also the docstring's motivating example (two triplet 3s → 3/3) cannot happen — 3/3 is rejected as invalid before the guard runs; *unequal* tuplet digits (3/4, 6/8…) are the real hazard and go unmentioned.

### Also wanted

- The sounding-glyph clamp is band-limited to the staff ±1 space and never documented as such: ledger-line notes — for guitar in treble-8vb, the open E/A/D strings — never clamp the window. Same blindness in the mid-system guard, which is additionally blind on the up-stem side (the stem's own notehead sits left of the barline and is invisible).
- "241 scores untouched" does not reproduce under any of nine definitions; the partition over the 293 extractable is **249 untouched / 38 better / 1 worse / 5 net-neutral** — and the 5 net-neutrals deserve a look (`Free from Fear`: inferred silence 1.0→5.0 q).
- The aggregate hides that inferred silence **rose on 14 scores by 117.5 q**, concentrated in headline wins (`A Better World` 39.75→81.0 q, 27 of 35 bars padded). Mechanically honest — marked, excluded from Rule 8 — but the PR body should say part of the defective improvement is bought with invented silence.
- Latent, each currently unreachable but wanting a pin: the `anchor_y` fallback makes `_ts_at`'s tuple comparison ignore x entirely for a low anchor (reverts #104 for unpaired tab staves, of which there are currently 0); `first_staff` is the first notation staff on the first *tab* page, not the score's first (relocates the backdating bug for tab-first scores, currently 0); `_ts_at` silently requires its timeline sorted and the new fixture test happens to pass unsorted input.
- The "opening not read" warning fires even when the later meter equals the assumption, interpolating the internal `ts_source` token into nested parentheses. Suppress or reword when there is no difference to report.

Issue #116, opened 2026-08-22
A unison shared by two voices binds both copies to one stem, and the other voice loses its note


sirged:
Found incidentally by the adversarial review of #108 while it was building a population count. Not related to that PR, and not investigated further there.

**2,881 noteheads in the library are drawn twice at the same position** — identical font family, glyph id, and x/y. The decoder treats them as two separate noteheads and **emits two notes for each**.

Duplicate printing is normal in engraved output: a glyph can be stroked twice for weight, or a page can overlay content. The question is whether the decoder should be collapsing coincident noteheads before reading rhythm. If it should, roughly 2,881 spurious notes are reaching the output, and each one either doubles a chord member or occupies a beat position that has nothing in it on the page.

The review noted this specifically because deduplicating them changed its own population counts by thousands, and it nearly drew a false conclusion from the difference — so this also affects any future measurement over noteheads.

### What needs establishing first

Whether these are genuinely the same notehead drawn twice, or two real coincident noteheads (two voices at the same pitch in the same place, which is legitimate engraving and *should* produce two notes). That distinction decides whether this is a defect or correct behaviour, and it cannot be settled from the coordinates alone — it needs a handful checked against the printed page.

If they are duplicates, a fix should collapse them and the sounding-note total should fall by close to 2,881 with no change to bar conformance figures. If bar arithmetic *improves* when they are collapsed, that is strong evidence they were spurious.

sirged:
## Human adjudication — the verdict is mixed, so a blanket dedup is wrong

Three examples were checked against rendered pages by a guitarist. The three deliberately spanned the heaviest score, a middling one, and a light one:

1. **Carulli (PrimoGuitar, 222 duplicates — the heaviest): artifacts.** The music is single notes throughout — "not duos or trios," all quarters and eighths. There are no unisons for a second copy to legitimately be, so every duplicated notehead there is the same head drawn twice. The three PrimoGuitar scores at the top of the count (222 + 198 + 180 — over half of all duplicates) share the same publisher, so this is very likely an engraving-toolchain habit of that source.
2. **Tarrega: fine** — nothing anomalous at the flagged spot.
3. **Born a Stranger (Oeth): legitimate.** Two notes adjacent on the same row — the lower drawn left-facing, the higher swapped right-facing on the same line — standard engraving for adjacent simultaneous notes. Two emitted notes is correct there.

[...]

Issue #112, opened 2026-08-22
A dot one space from its notehead is dropped, so a seconds-interval chord loses a dot


sirged:
Found while attacking the dot rules on #108, present identically there and in `main`.

Two notes a second apart cannot share a notehead column, so engravers offset the lower notehead to the left. When both are dotted, two things then break:

**The offset notehead is out of reach.** Its right edge ends up about 1.2 spaces from its dot, outside `dot_x_tol` (1.17 spaces), so the lower note of the interval silently loses its dot. Reproduced on an engraved fixture: E4 and F4 as dotted halves, the lower one reading undotted.

**The dot is placed further out than the tier ceiling allows.** Measured on that fixture, the lower dot sits a full space from its notehead (149.16 against a notehead at 146.69, spacing 4.975), which is beyond the +/-0.75-space ceiling in `_DOT_TIERS`.

That second point matters for how this gets fixed. #108 argues the tier ceiling cannot be widened, because a wider window would make the tiers overlap and leave nothing to rank on. That argument is sound as far as it goes, so this needs something other than a wider window — most likely knowing which notehead is the offset one and measuring from the column the dot belongs to rather than from the notehead's own edge.

Related: #111 covers a separate reach failure in the same rule.

### Verifying a fix

Both notes of a dotted seconds interval read `dots=1`. As with the double-dot case, dot counts on ordinary dotted notes elsewhere in the library must not move.

Issue #111, opened 2026-08-22
A genuine double dot reads as a single dot, losing half a beat


sirged:
Found while attacking the dot rules on #108. Present identically in that branch and in `main`, so it is pre-existing and not a regression.

An engraved fixture containing a **double-dotted half note** — two dots side by side, confirmed visually on the rendered page — decodes as `dots=1`. A double-dotted half is 3.5 beats; reading it as single-dotted makes it 3, losing half a quarter and putting every following onset in the bar early.

The mechanism is that `dot_x_tol` (1.17 spaces) is measured from the notehead's right edge, so it can only ever reach the *first* dot:

```
notehead x1 = 323.39
dot 1  xc = 326.85   window [321.03, 327.85]  inside
dot 2  xc = 330.08   window [324.26, 331.08]  outside
```

The second dot of a double dot sits roughly one dot-spacing beyond the first, which is always outside a window anchored on the notehead. Any fix needs the window for dot *n+1* to be anchored on dot *n* rather than on the notehead.

Worth noting for whoever picks this up: #108 rewrites the dot rule and argues at some length about double dots, but its reasoning is about ranking dots between competing noteheads, not about reaching the second dot at all. The two concerns are separate.

### Verifying a fix

The fixture is reproducible with the repository's own `engrave_fixtures` helpers. A fix should read `dots=2` on a double-dotted note and must not raise the single-dot count anywhere in the library — a change in dot counts on ordinary dotted notes would mean the window widened rather than chained.

PR #130, opened 2026-08-24
Bind each augmentation dot to its expected offset, not the nearest notehead (#89)

### LINKED ISSUE #89: Augmentation dots bind to the wrong notehead in a stacked chord
An augmentation dot is engraved slightly *above* the notehead it belongs to — roughly 0.58 staff spaces, which the dot-assignment code's own comment states. But each dot's owner is chosen by smallest vertical distance. In a chord of noteheads a space apart, that nudge means every dot is closer to the notehead **one step above** its true owner, so ownership shifts by one all the way up the chord and the end note collects two.

## Measured on main

Across 104,448 emitted notes: **878 double-dotted notes**.

A double dot is genuinely rare in this repertoire — it means "three quarters again as long", and fingerstyle guitar arrangements use it sparingly. Not 878 times. The count is the signature of misattribution, not of the music.

Examples: Score A, Score B, Score C.

## The concrete shape

A chord of three dotted half notes, noteheads one space apart, each with its own dot. Nearest-distance assignment gives the three noteheads 0, 1 and 2 dots, so three identical dotted halves decode as:

- a plain half — 2 quarters
- a dotted half — 3 quarters, correct by luck
- a double-dotted half — 3.5 quarters

One voice in that chord is short, another long, and the bar's arithmetic is wrong in both directions at once. Since a chord's notes are simultaneous by definition, the error is not recoverable downstream — nothing later in the pipeline knows the three should have been equal.

## The fix

Assign each dot to the notehead at the **expected offset**, not the nearest one. The offset is already known and documented in the code; using it turns the ambiguity into a match. A dot with no notehead at the expected position is then a genuine anomaly worth reporting rather than something to attach to the closest candidate.

Worth pairing with #88, since both are cases of comparing vertical positions without accounting for a known constant offset — and both were found the same way, by computing an invariant the music cannot violate and seeing it violated.

## How this was found, because the method transfers

Not by reading the code. By asking what the emitted output should never contain — a double dot at that density, a whole rest longer than its bar — and then counting. Both counts came back non-zero. That technique costs almost nothing and does not depend on guessing where a bug might be, which makes it worth reaching for on any decoder whose output has arithmetic constraints.

sirged:
Closes #89.

## The defect

An augmentation dot sits roughly half a staff space off the notehead it
belongs to - above it by default, or below/in its own space when the
engraver needed the room elsewhere (see `_DOT_TIERS`, from #108). Ownership
was still resolved by ranking each dot against every candidate notehead
**independently**: once a chord has three or more notes close together, one
notehead already given a dot at one relative position (say, "the space
above it") could still win a SECOND, different-position dot ranked against
it in isolation ("the space below it") - because nothing remembered that
the first dot had already spoken for that note. The result is one note
double-dotted and another with none, in the same beat.

**Note on the numbers in #89.** The issue's figures (878 double-dotted
notes, 104,448 emitted notes) predate #108's ink-centre fix and do not
reproduce on current `main` (8101b00) - measured there: 98,688 notes,
**919** double-dotted. That is the before-column used throughout this PR.
Two of the issue's three named scores (Score B, Score C) already read 0 double-dotted notes on current `main` for the same reason. **Score A** still has the defect and is verified below.

## The fix

`_assign_dots` now settles ownership by elimination instead of ranking each
dot alone:

- A dot reachable by only one candidate owner is bound to it immediately.
- Binding an owner to a tier removes that owner from competing for a
  DIFFERENT tier on any other dot - one note cannot be in two relative
  positions to its dot at once - which can turn what looked like a tie for
  another dot into a forced assignment too, so this proceeds by repeated
  elimination (least-ambiguous first).
- An owner MAY still take a second dot at the SAME tier it already holds:
  that is what a genuine double or triple dot is, two ink marks side by
  side, and #108's existing double-dot support is unaffected (5 committed +
  new unit tests all still pass).
- Only once elimination stalls does anything fall back to the old
  (tier, deviation, x-distance) ranking - now used to break a genuine tie
  rather than to decide everything.
- A dot no owner ever reaches, even after everything else has settled, is
  counted (`dots_unassigned`) and surfaced as a warning - the "genuine
  anomaly, not the nearest candidate" behaviour #89 asks for - instead of
  being silently dropped (what `main` does) or forced onto whichever
  notehead happens to be nearest.

One more thing this exposed, found on a real library score and fixed the
same way: two GlyphEvents (a notehead and its dot) can be recorded twice at
*identical* coordinates - some kind of duplicate content-stream artifact.
`_DOT_X_DUP_TOL` refuses a second dot at an x indistinguishable from one
already bound to the same owner. **What this constant actually enforces,
and its real weight in the fix, is corrected below (review finding 1) - the
paragraph above understates it.**

## #111 / #112

Neither is touched: `_bounds`, `dot_x_tol`, and `dot_x_back` - the x-window
mechanics both issues are about - are unchanged. Reachability is exactly
what it was; only the RESOLUTION among already-reachable candidates
changed. Measured no change in which dots fall inside or outside the
x-window anywhere in the library.

One of the two engraved-fixture cases below happens to sit on both issues
at once (a #112-style notehead offset orphans a dot, and old `main`'s
lack of exclusivity then lets that orphaning cause a wrongful double-dot
elsewhere) - this fix does not reconnect the orphaned dot to its own
notehead (#112 unfixed, unworsened), but it does stop that orphaning from
also corrupting a neighbour's count, which #108's original tier ranking
alone did not prevent.

## Falsifiable targets, measured

**1. Double-dotted notes: 919 -> 11.** All eleven are in one score,
`Sagreras-Book2-No5-Guitar.pdf` (a classical study, not fingerstyle guitar -
genuinely written with repeated double-dotted rhythms). Confirmed against
the rendered page: each is two ink marks side by side after the notehead,
not a decoding artifact.

```
                                     before   after
double-dotted notes                    919      11
single-dotted notes (notes only)      7724    8632
```

(The table originally posted here read 7781 -> 8689 on both sides - 57 of
what it counted on each side are dotted **rests**, not notes; corrected -
see review finding 5.)

**1a. Which clause does the work - the ablation, run both ways.** The
elimination machinery (least-ambiguous-first, fall back to ranking on a
genuine tie) is necessary but not sufficient on its own; the two
`_drop_conflicting` clauses are where the fix actually lives, and unevenly:

| variant | double-dotted | defective | overfull |
|---|---:|---:|---:|
| base (`main`, 8101b00) | 919 | 5477 | 1754 |
| elimination loop, neither clause | 919 | 5477 | 1754 |
| tier-exclusivity only | 13 | 5333 | 1578 |
| same-x dedup only | 11 | 5332 | 1577 |
| both (shipped) | 11 | 5332 | 1577 |

The propagation machinery alone reproduces the base **to the digit** - zero
eliminations satisfy only the tier predicate without help. Tier-exclusivity
is the principled rule (one note, one relative position) and is the one
that generalises; `_DOT_X_DUP_TOL` supplies the last two, and see finding 1
below for what it is actually doing while it does that.

**2. The three named scores.** Score B and Score C already read 0 double-dotted notes on current `main` (issue's figures
are stale - see above). **Score A** measures 5 and 9 (identical two-bar phrase): a G3+D3 half-note chord (voice 2, a fifth
apart) decoded as double-dotted (`dots=2`, 3.5 quarters) on `main`; with
the fix, both notes correctly read `dots=1` (3 quarters). Checked against
the printed page: G3 and D3 each carry exactly one augmentation dot, at
its own height beside its own notehead - `main`'s `dots=2` reading is a
decoding error, not something printed twice on the page.

**3. Aggregates, re-derived on this base** (before = measured on 8101b00,
matching the task's own expectation of "defective near 5,477"):

|                          | before (8101b00) | after  |
|--------------------------|-----------------:|-------:|
| bars                     |            10946 |  10946 |
| bars_defective           |             5477 |   5332 |
| bars_overfull            |             1754 |   1577 |
| bars_short               |             4193 |   4197 |
| bars_padded              |             3563 |   3567 |
| inferred_rest_quarters   |          4835.75 |4860.75 |
| sounding notes           |            98688 |  98688 |

Per-file: 233 of 293 extractable scores unchanged, 58 improved (148 bars
fewer defective between them), 2 worse (3 bars between them: `My Lady`
+1, `The Riverlands` +2). Both "worse" cases are the same shape, checked
bar by bar: a chord voice was previously wrong in BOTH directions at once
- one notehead double-counted a dot that inflated the voice to exactly the
bar's budget by coincidence - and reads as correctly short (padded with
inferred silence, not lost) once the dot count is right. Correcting a
compensating error can make a bar's arithmetic look worse; the note
durations underneath it are more correct either way.

New library-wide honesty metric this PR adds: **1,270 dots_unassigned**
across 520 staff decodes in 206 of 293 scores - dots this decoder now
admits it cannot place, mostly at the same x-reach limits #111/#112
describe, rather than silently forcing them somewhere. No comparable
"before" figure exists; `main` never counted this at all. **Now split into
1,217 that never had a candidate at all and 53 that reached one and lost
it during elimination - see review finding 3 below for why the split
exists and what each half actually claims.**

**4. Sounding notes: unchanged (98,688 -> 98,688).** Nothing here reads a
pitch differently, only how many dots attach to it.

**5. Engraved fixture** (`stacked_dotted_chord`): a three-note chord, E4 +
G4 + A4, every member an identical dotted half. G4 sits a second below A4,
close enough that this engraver offsets G4's own notehead left of the
shared stem (the #112 shape) - which also carries G4's own dot glyph
outside this decoder's x-reach entirely. That is a genuine orphaned dot on
its own, not this issue's defect. The defect is what `main` does to A4's
OWN dot with G4's no longer in the running: nothing else is near enough to
rank it against, so A4 takes it, and separately takes the next dot along
too - two dots at two different relative positions, which is not a real
double dot. Measured directly off `decode_note_events` (before the beat
model recombines the chord): `main` gives A4 `dotted=2`, G4 `dotted=0`,
E4 `dotted=1` - the chord reads double-dotted (3.5 quarters) instead of
the 3 every member is actually written as. With the fix: A4 `dotted=1`,
G4 `dotted=0` (still orphaned - #112 is not this fix's job), E4
`dotted=1`, and `dots_unassigned` counts G4's dot once per bar instead of
it vanishing into A4's total. (G4's dot glyph, though off its own
notehead's x-reach, still falls inside A4's window and is dropped there
once A4 is already committed - so it is the *eliminated* half of the
split, not the *no-candidate* half: `dots_unassigned_eliminated == 7`,
`dots_unassigned_no_candidate == 0` for this fixture.)

## Tests

- `test_a_three_note_chord_gives_one_dot_to_each_notehead` (unit, built
  from the real coordinates of the affected chord in `Courage`): 0/1/2 on
  the old ranking, 1/1/1 with the fix.
- `test_a_stacked_chords_dot_does_not_bind_to_the_nearest_notehead`
  (engraved fixture, real PDF geometry): the red/green case above, plus
  `dots_unassigned` (and now its split) firing on real/engraved input
  rather than only a synthetic branch.
- `test_an_eliminated_dot_is_distinct_from_one_with_no_candidate` (unit,
  new): a synthetic case built specifically to exercise the *eliminated*
  half of the split in isolation from the *no-candidate* half.
- The 5 existing dot tests (one-dot-one-owner, genuine double dot, stacked
  chord of two, dot-below, dot-a-whole-space-away) all still pass unchanged
  in behaviour, updated only for the new `(counts, unassigned_no_candidate,
  unassigned_eliminated)` return shape.
- `test_unassigned_dots_survive_the_api_round_trip` (new, real library
  fixture): `dots_unassigned` / `dots_unassigned_no_candidate` /
  `dots_unassigned_eliminated` / `staves_dots_unassigned` all survive a
  POST -> GET round trip, non-zero, mirroring `notes_no_stem`'s coverage -
  see review finding 4.

**Proved able to fail**, each reverted separately, full suite green
otherwise:

| mutation | red |
| --- | --- |
| `_assign_dots` reverted to 8101b00 (nearest-independent ranking) | the new unit test AND the new engraved-fixture test - `top.dotted`/`A4.dotted` read `2`, not `1` |
| the `unassigned += 1` line removed (anomaly counter broken) | `test_a_dot_a_whole_space_from_every_note_belongs_to_none_of_them` AND the engraved-fixture test's `dots_unassigned` assertion |
| both `_drop_conflicting` clauses disabled at once (the ablation table's "neither clause" row) | `test_a_three_note_chord_gives_one_dot_to_each_notehead`, the engraved-fixture test, AND `test_an_eliminated_dot_is_distinct_from_one_with_no_candidate` |

(The table originally posted here listed 2 rows; the third was measured
and is added here - review finding 5.)

## Review round two - findings applied

A second pass ("correct on every attack, wrong in its account of itself")
re-verified the algorithm itself - order independence, convergence, the
tier lock - and found it sound; every item below is documentation,
counter-shape, or test-gap, not a behaviour change. #131 (a genuine
double-dotted note inside a stacked chord still sending both dots to the
wrong member) is filed separately and out of scope here.

**1. `_DOT_X_DUP_TOL`'s comment was wrong by a factor of ~280.** It called
itself a guard against "an artifact seen once in the library and unrelated
to dot ranking." Measured: the clause removes 567 candidates across 136
scores, of which exactly 2 are that artifact. The other 565 are two
genuinely distinct dots one space apart in the same x column - two chord
members each dotted. What it actually implements is "one owner takes at
most one dot per x column" - stronger than deduplication, and still
correct (the real second-dot gaps in the library's double-dot fixtures run
~0.5 spaces, ten times the 0.05 tolerance) - but as documented, a future
#116 cleanup would have read it as dead artifact-guarding and deleted it,
silently restoring most of the 919. Rewritten in place. Also newly
documented: dropping a candidate here (rather than never having offered
it) can redirect a true duplicate's dot onto an unrelated, dotless
neighbour instead of reporting it - an error class the old
nearest-independent ranking could not make. About 2 possible instances
library-wide, not traced further.

**2. The ablation table** is now in this body (section 1a above) instead
of only in review comments.

**3. `dots_unassigned` conflated two facts.** "No candidate ever" (1,217)
and "eliminated by this algorithm" (53) are different claims about the
page - the first says no notehead or rest sat anywhere near the expected
offset; the second says one did, and had already been given its own dot
elsewhere - while four docstrings and the warning prose asserted the first,
stronger claim unconditionally. Split at the source:
`_assign_dots` now returns `(counts, unassigned_no_candidate,
unassigned_eliminated)`, summed unchanged into the existing
`dots_unassigned` everywhere that field already flows (`ExtractionResult`,
the stored confidence blob, `_BAR_KEYS`) - **the combined figure is kept,
not migrated**, since nothing currently consuming it needed to change, and
two more precise fields sit beside it for anything that wants the
breakdown. Every docstring and the warning sentence itself now say which
is which instead of the single overclaiming sentence.

**4. Test gaps against the field's own model.** `dots_unassigned` /
`staves_dots_unassigned` were added to the real `_BAR_KEYS` tuple in
`api.py` when they were introduced, but never added to the **hand-copied**
`BAR_KEYS` tuple test-side (`test_transcription_api.py:370`) - so nothing
ever exercised their reload path, the same gap `notes_no_stem` had before
#115's own test was added. Added, along with the two new split keys, plus
a real-library round-trip test
(`test_unassigned_dots_survive_the_api_round_trip`) mirroring
`notes_no_stem`'s non-zero coverage. Proved red first: hardcoding
`dots_unassigned` to `0` in the API layer's confidence blob turns that new
test red (`assert 0 == 16`); reverted before this PR.

**5. Smaller.**
- The dedup-redirect error class above: documented, not fixed (no known
  instances beyond the ~2 estimated).
- `while _drop_conflicting(): pass` can never run a second pass in the same
  outer iteration: the predicate reads only `owner_tier` /
  `owner_dot_xs`, and neither changes except inside `_commit`. Reduced to a
  single call; verified by instrumenting a second, do-nothing call
  immediately after it and running the whole library through it - 0 of
  12,631 calls ever found anything further to drop.
- `engrave_fixtures.py`'s `mirror_to_tab` had a dead
  `.replace("<voice>3</voice>", "<voice>7</voice>")` - no fixture ever
  generates voice 3 (only 1 and 2). Removed.
- The single-dot figure labels are fixed above (section 1): 57 of what was
  counted as notes on each side are dotted rests.
- The ranking key's `round(deviation, 3)` / `round(xdist, 3)`, present in
  the pre-fix single-dot ranking, had been silently dropped by the
  elimination rewrite. Restored. Re-measured against the whole library with
  and without it: identical (11 / 5,332 / 1,577 / 98,688 either way).
- The revert-mutation table listed 2 rows; the actual number tested is 3 -
  added above.

## Not verified

- Whether the duplicated notehead+dot glyph pair `_DOT_X_DUP_TOL` guards
  against (2 of the 567 candidates it removes) is a one-off artifact of
  that single PDF or a recurring pattern worth its own issue - not traced
  to a cause beyond "the content stream draws it twice." What the clause is
  actually doing the REST of the time (567 - 2 = 565 candidates, "one owner,
  one dot per x column") is now measured and documented, not merely
  asserted - see finding 1.
- The ~2 possible instances of the dedup-redirect error class (finding 1)
  are an estimate from the measured drop count, not individually located
  and confirmed against the printed page the way the double-dot cases
  above were.


sirged:
## Review — correct on every attack, wrong in its account of itself; fixes wanted before merge

The algorithm survived everything aimed at its structure: order dependence (every multi-dot call in the library re-run with the dot list reversed — 0 of 1,829 change; exhaustive permutation of synthetic cases — one outcome), convergence (each iteration resolves exactly one dot; the filter is monotone), the tier lock (no constructed wrong-first-binding makes the new code wrong where the old was right). Both "worse" scores were verified against the printed page and are the fix working — a spurious dot had been exactly covering a genuinely short bar. Three of the 11 surviving double dots were independently confirmed as real ink. The count is real and the partition (233/58/2) reproduces exactly.

### What must change before merge

**1. `_DOT_X_DUP_TOL`'s justification is wrong by a factor of ~280, and it is the load-bearing rule.** The comment says "an artifact seen once in the library and unrelated to dot ranking." Measured: the clause removes **567 candidates across 136 scores**, of which exactly **2** are true duplicates. The other 565 are two genuinely distinct dots one space apart in the same x column — two chord members each dotted. What the constant actually implements is "one owner takes at most one dot per x column," which is stronger and still correct — the 10 real second-dot gaps measure ~0.5 spaces against the 0.05 tolerance. But as documented, a future #116 cleanup would delete it as an obsolete artifact guard and silently restore most of the 919. Rewrite the comment and the PR's *Not verified* paragraph to say what it does and how often.

**2. The ablation, run both ways:**

| variant | double-dotted | defective | overfull |
|---|---|---|---|
| base | 919 | 5477 | 1754 |
| elimination loop, neither clause | **919** | **5477** | **1754** |
| tier-exclusivity only | 13 | 5333 | 1578 |
| same-x dedup only | 11 | 5332 | 1577 |
| both (shipped) | 11 | 5332 | 1577 |

The propagation machinery alone reproduces the base **to the digit**; zero eliminations satisfy only the tier predicate. The tier rule is the principled one and stays — but the PR body and the ~45 docstring lines must state the measured division of labour instead of implying the apparatus earns the number.

**3. `dots_unassigned` conflates two facts** — "no candidate ever" (1,217) and "eliminated by this algorithm" (53) — while four docstrings and the warning prose assert unconditionally that these dots had no notehead at the engraved position, untrue for 4% of the counter. Split the counter or soften every claim to match.

**4. Test gaps against the field's own model.** `notes_no_stem` is asserted at extraction level, in the API response, and across a reload; `dots_unassigned` has neither of the latter two. And `test_transcription_api.py:370`'s hand-copied `BAR_KEYS` tuple was not updated, so the reload test silently does not cover the two new keys. Add the keys and mirror the `notes_no_stem` assertions.

**5. Smaller:** the dedup can *redirect* a true duplicate's dot onto a dotless neighbour rather than report it (an error class the old code could not make; ~2 possible instances library-wide — document it in the comment); `while _drop_conflicting(): pass` can never iterate twice (the predicate reads only state the function never mutates — measured max passes 1 of 12,065; make it a single call); `engrave_fixtures.py:115`'s `voice>3` replace is dead; the single-dot figures label 57 dotted **rests** as notes (7,781/8,689 should be 7,724/8,632 for notes); the ranking key silently lost its `round(...,3)` — measured identical either way, but say so or restore it; the revert-mutation table lists 2 red where the measured number is 3.

### Recorded, not for this PR

A genuine double-dotted note in a stacked chord still sends both dots to the wrong member — structurally, since the same-tier exemption is exactly what lets the wrong owner take both. Filed separately.

Issue #134, opened 2026-08-24
Phase 2: D.C., D.S. and al Coda text instructions are not read (repeats and voltas landed in #138)


sirged:
Reported from real use: a library score plays measures 1-6, then the first ending, then 1-6 again, then the second ending — and the transcription ignores all of it, playing straight through linearly.

Verified on `main` (d1216a8): the emitted MusicXML for that score contains **zero** `<repeat>`, `<ending>`, coda, segno or even `<barline>` elements. The pipeline is not unaware of these symbols — the Finale glyph table maps the coda glyph, and the staff detector explicitly recognises the Oeth "1." / "2." volta brackets *in order to avoid mistaking them for staff lines* (`tabextract.py:398`) — but nothing decodes them into structure. Every mark of musical form on the page is discarded.

## Consequences, in order of player pain

1. **Playback plays the piece wrong.** Not a degraded approximation — structurally wrong, in a way any player following along notices in the first repeat. The one confirmed daily user found this unprompted.
2. **The transcription misrepresents the piece's length and shape.** 24 linear bars stand in for a form the page writes as 6 bars, two endings, and a return.
3. **Practice tooling built on bars inherits the wrong form** — loop points, position memory, the metronome's bar counting.

## Scope worth splitting

- **Repeat barlines and volta brackets** are the common case (the Oeth arrangements use them constantly) and are geometrically tractable: repeat dots are already in the glyph stream, thick/thin barline pairs are vector lines, and the volta brackets are already being detected and thrown away.
- **D.C./D.S./al Coda text instructions** are text-layer reads, same machinery as the tuning labels (#80) — a separate, later chunk.
- **MusicXML has first-class representation for all of it** (`<barline>` with `<repeat>` and `<ending>`, direction elements for segno/coda), and alphaTab honours repeat structure in playback, so the whole fix is on the extraction side.

## Verifying a fix

That score plays 1-6, ending 1, 1-6, ending 2 — confirmed by the same person who reported it. The emitted MusicXML carries the repeat barline and both endings; a MusicXML consumer (MuseScore round-trip) preserves them. The library-wide count of scores gaining repeat structure should be large — most of the Oeth arrangements carry at least one repeat — and bar-conformance figures must not move at all, since form marks carry no duration.

sirged:
## Design specification: reading repeat structure off the page

Measured over the whole library (297 PDFs, 833 pages, 2,786 tab staves) plus the
[...]

PR #138, opened 2026-08-27
Read repeat barlines and volta brackets (#134 phase 1)

sirged:
## Summary

Phase 1 of #134, in two commits:

1. **`_detect_barlines` merges a repeat pair's two strokes into one boundary.** A repeat barline is drawn as a thin stroke plus a thick stroke a few points apart; the previous 2.0pt merge tolerance left both as separate barlines with a phantom sliver "measure" between them - 314 phantom measures across 169 of 297 scores, all counted as bars nothing was read from. Merging within 1.0 staff space (the empty 0.526-2.273 space band measured across 12,228 consecutive-vertical gaps) removes them. Library-wide: `bars` 10,946 → 10,632, `bars_unread` 337 → 23, with `bars_overfull`/`bars_short`/`bars_defective`/`bars_padded`/`inferred_rest_quarters`/`notes` identical to the unit before and after.

2. **Repeat barlines and volta brackets are read and written as `<barline>`.** The stroke width `_vertical_segments` discarded now travels with each segment, so `_detect_barlines` can classify a merged group's thin/thick shape and search for repeat dots (`glyph_rhythm.dot_like_glyph_events` does the glyph classification; the geometry stays in `tabextract`). Volta brackets are read fresh from the drawing primitives - the staff detector's own line-welding pass discards that geometry on purpose - anchored above the system's topmost staff by a downward hook landing on a barline, with the ending's extent taken from the bar grid and a closing backward repeat rather than the bracket's own drawn (and sometimes very displaced) right end. A partial read is omitted and disclosed via new `ExtractionResult` fields and a `structure` confidence key, never guessed. `musicxml.build()` gains a `barlines` parameter; a form mark carries no duration and cannot move a Rule 8 figure.

Codas, segnos and text navigation (`D.C.`, `D.S.`, `Fine`, `al Coda`) are explicitly out of scope for phase 1 - no segno glyph is drawn anywhere in the library, and alphaTab's MusicXML importer only reads a jump from a `<sound>` that is a direct child of `<measure>`, not the nested form every notation program writes.

## Targets measured

- **Commit 1:** `bars` 10,946 → 10,632 (−314), `bars_unread` 337 → 23 (−314), `beats` −408, `notes`/`bars_overfull`/`bars_short`/`bars_defective`/`bars_padded`/`inferred_rest_quarters` unchanged to the unit.
- **Score A (the acceptance case):** 23 bars, forward repeat at measure 1, ending 1 over measures 7-8 closed by a backward repeat, ending 2 at measure 9 left open, `<repeat>`/`<ending>` matching the hand-verified structure exactly, and alphaTab playback order `1-8, 1-6, 9-23`.
- **Library-wide:** 188 of 297 scores gain repeat structure; conformance figures (`bars_overfull`/`short`/`defective`/`padded`/`inferred_rest_quarters`/`notes`) unmoved by commit 2, verified to the unit.
- Suite: 720 passed/2 skipped on `ee43e1c` baseline (with `FERMATA_TEST_LIBRARY` + `npm ci`); 725 passed/2 skipped on this branch (5 new tests).
- Both required mutations proven red then reverted: dropping the stroke-width read collapses every repeat/volta detection; restoring the 2.0pt merge tolerance reintroduces the phantom bars.

## Test plan

- [x] `docs/musicxml-tab-profile.md`'s Rule 14 figure updated (338 → 24 measures of rests) and new Rule 15 section added
- [x] `fixture_repeat_structure()` added to `engrave_fixtures.py` with `barline()`/`repeat_forward()`/`repeat_backward()`/`ending()` helpers; `volta.pdf` left byte-identical
- [x] `verify_musicxml.mjs --repeats` reports `isRepeatStart`/`repeatCount`/`alternateEndings`/`tickLookup`, asserted against the real alphaTab importer
- [x] Score A and full-library tests gated on `FERMATA_TEST_LIBRARY`
- [x] Both red-condition mutations manually proven and reverted (see above)
- [x] `engrave_fixtures.py --check` passes

## Adversarial review follow-up

Every finding from the adversarial review addressed, in the review's own order. Figures re-derived directly against the current library (293 extractable of 297 PDFs) - the review's own census (175 brackets) does not match this run (247 numbered brackets today), which is explained below rather than force-fit.

**Blocker 1 - volta anchor guard order.** `_associate_voltas`'s `nearest_barline` guard ran before `_anchor_mark`, rejecting every bracket anchored via `_anchor_mark`'s case 2/3 (a system-start/end bracket with no real barline stroke at its x) before that anchoring was ever reached. Moved the guard after anchoring, applied only to case 1. `_anchor_mark`'s own case 4 (no boundary within reach) was structurally dead - three exhaustive branches - fixed with a real proximity test keyed to the tab staff's own spacing (not the notation staff's, which was too tight and broke Score A's own second ending during development - see commit history).
- Emitted `<ending type="start">`: **164 → 242** (library holds 247 numbered brackets today, not the review's 175; 242/247 = 98% conversion, the 5 residual are `endings_unread`, disclosed).
- Scores emitting ending 2 with no ending 1: **14 → 2**, not 0. Both residuals independently verified as correct, not new bugs: *Ahead on Our Way* has a genuinely unreadable ending-1 number (now disclosed via `endings_unread`, bar 7); *Score B*'s only "1"-shaped mark sits 1.93 std-spaces from the nearest barline (vs. the calibrated max of 0.491), correctly rejected by the same discriminator that rejects a ledger line - both now show `endings_incomplete=1` and a downgraded `structure` confidence.
- Score C (the review's own acid test, p1 sys3) verified directly: both endings now read correctly, exact structure pinned in a new test.
- `endings_unread`: **0 → 5** across 4 scores (was structurally unreachable before the fix).
- `form_marks_unanchored`: stays 0 (no numbered bracket in the library hits `_anchor_mark`'s case 4), but the mechanism is now real - proven via a direct unit test and 200k-probe-style reasoning, not a tautological constant.

**Blocker 2 - confidence sum.** `structure_issues` omitted `endings_incomplete`. Fixed; isolated with a new library-gated test (*Score B*, where `endings_incomplete=1` is the *only* nonzero structure figure).

**Blocker 3 - API allowlists.** The nine repeat/volta disclosure fields reached `to_dict()` but nowhere in `server/` calls it; `api.py`'s `_BAR_KEYS`/`_BAR_LIST_KEYS` and the `POST /transcribe` confidence blob never named them. Added to both; round-trip test added mirroring `notes_no_stem`'s coverage, against a fixture with a genuinely nonzero count (Tarrega-Estudio-Em, `repeats_unread=1`).

**Items 4-6 - unreachable branches.**
- Item 4: implemented the 0.5-space-scale proximity test for `_anchor_mark` case 1 (calibrated to 1.5 tab-staff-spaces after the Score A regression above); case 4 is now reachable, proven by a direct unit test.
- Item 5: repeat-dot search no longer gated on `"H" in shape` - a missing stroke `width` (substituted 0.0) read a genuinely thick stroke as thin and silently dropped its repeat with zero disclosure. Proven with a synthetic `_detect_barlines` unit test (real library fixture for this exact combination not found).
- Item 6: two-or-more thick strokes with no repeat direction resolved now emit `heavy-heavy` with disclosure instead of nothing at all. Real fixture: Tarrega-Estudio-Em ("tHHt", bar 8) - now `repeats_unread=1`, both sides of the boundary carry `heavy-heavy`.

**Item 7 - silent node_modules skips.** Nine tests (not just the Score A/playback headline pair) skip without `web/node_modules`. Counted and announced in the terminal summary the same way library skips already are, with its own unit test.

**Item 8 - XSD coverage.** `_emitted_samples()` never passed `barlines=`; docs carried no barline example. Added a `"barlines"` sample (forward repeat, an ending closed by a backward repeat, one left open) to both `test_musicxml.py` and a new published `docs/examples/repeat-structure.musicxml`, both validated against the real MusicXML 4.0 XSD in CI.

**Item 9 - `stop` vs `discontinue` at an abutting bracket.** Built a new engraved fixture (`fixture_adjacent_endings`, MuseScore 4.6.3) with a discontinued ending immediately followed by the next bracket's own opening hook, no bar in between - the case neither committed fixture reached. It surfaced two real, previously-invisible bugs:
- `has_right_hook` excluded only the *current* bracket's own left hook, so the neighbour's opening hook was read as this bracket's own closing one. Fixed with a hook-occurrence count compared against how many brackets claim that x as their own left hook (not a flat exclusion - Score A draws two *coincident* hook strokes where one ending's real closing hook lands exactly where the next one's opening hook does, and that legitimate case still has to work).
- The backward-repeat search that resolves an ending's last bar had no upper bound and could walk into the *next* bracket's own backward repeat, silently overwriting that bracket's own closing mark. Bounded the search at the next bracket's own first bar.

Both fixed; both proven red via mutation, then restored. Neither moves the emitted `<ending>` count or any library-wide figure - invisible on every existing fixture and every library score sampled.

**Smaller findings**: dead duplicate `_append_barline` in `musicxml.py` deleted; `_VOLTA_NUMBER_RE` now rejects `"0"` and leading zeros (MusicXML's own `ending-number` pattern); `BARLINE_STROKE_MERGE_SPACES`'s "empty band" comment scoped to tab staves, with the notation-side margin re-measured and pinned (31,060 gaps sampled, 3,410 inside the tab-calibrated band, worst hop 0.995 against the 1.0 threshold - not the review's 1,206, the library differs, but the worst-hop figure matches almost exactly); `endings_truncated` pinned in the library-wide scan (25 across 22 scores, matching the review's own finding); `_detect_barlines`'s `page=None` docstring claim corrected (shape can still hold `"H"` with no page).

## Conformance invariant, held

All six Rule 8 figures re-derived directly against the library and pinned in `test_library_wide_repeat_structure_leaves_conformance_untouched` (now includes `inferred_rest_quarters`, previously tracked but not asserted): `bars_overfull=1569`, `bars_short=4219`, `bars_defective=5344`, `bars_padded=3589`, `inferred_rest_quarters=4885.0`, `notes=98688` - identical before and after every change in this round. Voltas and repeats carry no duration; none of this round's fixes could move them, and none did.

Suite: 736 passed / 0 skipped locally with `FERMATA_TEST_LIBRARY` + `npm ci` + `FERMATA_MUSICXML_XSD` set (11 new tests over the 725-passed baseline this branch started this round at). Every fix mutation-tested: reverted, confirmed red, restored, confirmed green.


sirged:
## Adversarial review — the core holds; the volta association layer does not ship as-is

**What survived attack, verified independently:** the conformance invariant holds exactly (all 297 scores, all six figures identical to the unit at base and both commits, 0 scores moved individually); the thick/thin classifier's gap is genuinely empty (49,474 strokes, nothing between 0.906 and 2.268pt); the chaining cannot bridge two real barlines (max hop 0.534 spaces on tab staves); repeat-dot reading is position-perfect with one already-disclosed exception; the emitted XML validates against the real MusicXML 4.0 XSD and survives a MuseScore round-trip losslessly; the playback assertion is genuinely red-able (linear input and swapped directions both fail it); Zelda's Lullaby reproduces the hand-verified structure exactly.

### Blockers

**1. Voltas beginning at a system start are silently dropped — a third of the library's brackets.** `tabextract.py:1156-1159`: the `nearest_barline > spacing * VOLTA_ANCHOR_SPACES` guard runs BEFORE `_anchor_mark`, the function that exists precisely for marks with no barline at their x (spec case 2, 37% of all marks). A bracket opening a system sits past the clef and key signature — 3.8 to 9.5 spaces from the system barline, clustering on discrete clef+key widths — and is rejected before the correct handling is reached. Measured: 164 endings emitted against the census's 175 brackets; **82 candidates across 59 scores** killed by this guard, reading only "1" or "2" (the signature of real voltas). **14 scores emit an ending 2 with no ending 1 anywhere** — verified on Lenna's Theme p1 sys3 against the page — which plays wrong in a NEW way. And `endings_unread` stays 0 because the counter only fires after the guard: zero disclosure. Move the guard after `_anchor_mark`, or fold the anchor cases into it; the 82 candidates and the 14 broken scores are the acceptance set.

**2. `structure` confidence says "high" on 12 of those 14 broken scores.** The `structure_issues` sum (~line 3915) omits `endings_incomplete`. A score simultaneously warns that its ending numbers do not form a run and reports "high - read directly from the engraving". Add the term; mirror how `rhythm` appends its defect count.

**3. None of the nine new disclosure fields reach a consumer.** `to_dict()` is never called in `server/`; the API's explicit field list and the `_BAR_KEYS`/`_BAR_LIST_KEYS` allowlists were not extended, so `repeats_unread`, `endings_unread`, `endings_truncated`, `form_marks_unanchored`, `endings_incomplete` and their `*_bars` lists reach nobody. The prose half works; the structured half is the recurring server-half-only defect by name. Add them to the allowlists, and extend the API round-trip test the way `notes_no_stem`'s coverage does.

### Unreachable branches and vacuous assertions

**4. `form_marks_unanchored` cannot be non-zero** — `_anchor_mark`'s three branches are exhaustive over the reals, the trailing `return None, None` is dead (0 hits in 200k randomized probes), and the spec's 0.5-space proximity test for case 1 is simply not implemented, so case 4 can never fire. The library test asserts a constant. Implement the proximity test, at which point the counter and its test become real.

**5. Spec row "dot pair but no thick stroke → disclose" is unreachable** — dots are only searched when a thick stroke was already found — and becomes reachable exactly when it matters: `_vertical_segments` substitutes 0.0 for a missing `width` (line 594, 34 such strokes in the library), a thick stroke reads thin, and the repeat is lost with no disclosure. Search dots regardless of shape, and route the mismatch to `repeats_unread`.

**6. Two thick strokes with unresolved dots emit nothing at all** — no bar-style, no warning (`continue` at 871-872; 2 instances, Tarrega-Estudio-Em). Emit heavy-heavy with disclosure instead of dropping the barline.

### Coverage gaps to close in this round

**7. Nine tests skip silently without `web/node_modules` — including the Zelda and playback-order headline tests.** The known trap, now three tests wider. At minimum update the documented count; better, make the skip loud in the terminal summary the way the library skips already are.

**8. The new elements never reach XSD validation** although CI sets the schema: `_emitted_samples()` never passes `barlines=` and the docs examples carry none. Add a barlines-bearing sample to both. (The review validated all four documents manually — they pass — but nothing enforces it.)

**9. `stop` vs `discontinue` where the next bracket abuts (43 of 164 endings): `has_right_hook` finds the NEXT bracket's opening hook.** The one page rendered happens to agree, and both fixtures declare adjacent endings as `stop`, so the assertion cannot distinguish the rule from luck. Add a fixture with an adjacent `discontinue` ending; if MuseScore cannot engrave the distinguishing case, document the ambiguity at the code site instead of resolving it silently.

### Smaller

`endings_truncated` measures 25 across 22 scores against the spec's predicted 3 — disclosed but unpinned; add the number to a library-gated test. The chaining comment's "empty band" claim is tab-staff-only — on notation staves 1,206 gaps land inside it with a worst hop of 0.995 against the 1.0 threshold; scope the comment and pin the notation-side margin. Dead duplicate `_append_barline` at `musicxml.py:416` (shadowed leftover) — delete. `_VOLTA_NUMBER_RE` accepts `0`/leading zeros which the XSD rejects — guard it. `_detect_barlines`' docstring claims `page=None` yields shape "t" — it doesn't; fix the claim.

PR #139, opened 2026-08-27
Let a unison inside a chord sound in both its voices (#137)

### LINKED ISSUE #137: Coincident duplicates inside a chord read as a spurious whole note on a library score
The one named residual from #136 (merged): after the unison split landed, a library score gained 12 defective bars. Page-checked during that work: its coincident duplicate noteheads sit inside a three-note chord, structurally different from the two-voice unison case the fix addresses, and the failure is a spurious whole-note duration rather than a different-onset borrow.

Scope: one score, +12 bars, visible in its conformance figures rather than hidden. Wants its own reproduction (the chord geometry is known from the #136 investigation) before any fix - and the fix must not regress the unison split, whose fixture and library assertions now pin the main behaviour.

sirged:
## Summary

A unison shared between two voices is **one plucked string**, so the tablature prints **one** fret number for it however many voices sound it. `_match_onset_columns` hands out an onset's digits one per notehead, ranked by pitch against string — right whenever every notehead is its own sounding note, and a digit short whenever one of them is a coincident duplicate.

That never surfaced under #116 because a unison **alone** does not run short: with nothing else at the onset the engraver is free to write each voice on its own string and does (MuseScore writes the `unison_voices` fixture's two voices as 2 on the fourth string and 7 on the fifth). Put the unison **inside a chord** and the freedom is gone — the chord's own members are what the column names — so the onset carries three noteheads over two digits, the third notehead got nothing, and the voice it belonged to lost its note for that beat.

**The fix.** Where a notehead is left with no fret number and a **coincident twin in another voice** has one, it takes the same digit — but only where the tab **fully named** the onset: `len(digits) == len(distinct notehead positions)`, with more than one position (i.e. a chord). Only then is the column demonstrably complete and the leftover copy provably a second stem on a string the tab already names.

**Why the chord condition is required and not cosmetic.** Without it the rule also fires where the *whole* onset is the coincident pair — one printed notehead over one printed digit. There the page is self-consistent as a single note, and the only thing suggesting otherwise is the two-opposing-stem signature #116 measured to be unreliable on its own. Its adjudicated example, **Carulli-Moderato-Op192** (reads as single notes on the printed page, carries that signature 74 times), went **308 → 382 notes and 0 → 2 defective bars** on the first, unguarded version of this fix. Library-wide the two families are 500 pair-alone onsets against 65 in-chord ones.

**The shared digit is disclosed, not silent.** A digit handed to a notehead the tablature printed no number for is an inference, so it is counted as `unison_digits_shared` and stated in a warning — the same convention as `notes_no_stem`, `dots_unassigned*` and `coincident_unsplit_pairs`. Carried through `ExtractionResult`, `to_dict()` **and** api.py's `_BAR_KEYS`, so the reload path has it too rather than the prose half of the disclosure reaching a reader while the data half reaches nobody.

## On #137's "spurious whole note" — corrected

An earlier draft of this PR said the `:16 3.1 :1 5.1` whole note **"does not reproduce."** That was wrong, and the claim is withdrawn. It reproduces at every commit checked. My method was narrower than the sentence I wrote: I diffed split-against-nosplit, saw the whole note in neither side of the diff, and generalised that into a claim about the score. It is present in *both* sides, and always was.

| commit | | bar 13 present | whole-note tokens | whole-REST tokens |
|---|---|---|---|---|
| `d07387d` | before #116 | **yes** | 3 | 2 |
| `ee43e1c` | #116, the unison stem split | **yes** | 3 | 2 |
| `16a33dd` | after #134 | **yes** | 3 | 0 |
| this branch | | **yes** | 3 | 0 |

The accurate statement: it is **Score A's bar 13, voice 1** — 0.25 + 4.0 = **4.25 quarters in 4/4**, one of the score's 8 overfull bars. It **predates #116 entirely**, is **not caused by the split**, is **not among the 12 bars fixed here** (those are 14, 16, 18, 20, 58, 60, 62, 64, 66, 68, 70, 72, all *short* bars), and is **not fixed here**. (The two whole *rests* that vanish at `16a33dd` are #134's doing, unrelated.)

Triaged rather than left hanging: rendering the site shows a **harmonic** — a grace note with a slashed sixteenth flag into a hollow notehead on a ledger line, `5` on string 1 — whose hollow head the decoder categorises `notehead_whole` (gid 84, no stem, no candidate stems) and emits at 4.0 quarters. That is the harmonic-notehead gap already named in the fixtures README, now with a measured cost. Filed as **#140** with the evidence.

Because #137's text named two defects, this PR closes it for the digit one and **#140 carries the whole note**, so nothing named in #137 closes untracked.

## Targets, measured

Method for every library figure: `tabextract.extract()` over all 297 PDFs of `FERMATA_TEST_LIBRARY` (293 extractable), aggregated exactly as `test_library_wide_repeat_structure_leaves_conformance_untouched` does, and additionally recorded **per score** so a delta is attributable rather than summed. "Before" is `16a33dd` extracted with `git archive` into a clean tree; "after" is this branch. Both under `py -3.13` on the same library.

### 1. Score A — the target score

| | before (`16a33dd`) | after | delta |
|---|---|---|---|
| bars | 78 | 78 | 0 |
| notes | 902 | **914** | **+12** |
| beats | 785 | 797 | +12 |
| bars_overfull | 8 | 8 | **0** (bar 13 is #140) |
| bars_short | 31 | **19** | **-12** |
| bars_defective | 35 | **23** | **-12** |
| bars_padded | — | — | -12 |
| inferred_rest_quarters | — | — | -6.0 |
| `unison_digits_shared` | 0 | **12** | +12 |

The issue's "+12 bars" reproduces exactly. Attribution measured, not assumed: neutralising #116's split alone (`_rank_stems` truncated to its head) gives 23 defective at this HEAD, so all 12 are the split's, and 23 is where this score sat before #116. **It does not merely return to 23** — it returns carrying 12 more notes, because the shared note now sounds in *both* its voices:

```
bar 14, upper voice, last beat:  :4 (1.2 00.3)     <- the chord keeps both members
bar 14, lower voice, 6th beat:   :8 00.3            <- and the lower voice sounds the same string
```

### 2. Library-wide aggregate (293 extractable of 297)

| | before | after | delta |
|---|---|---|---|
| bars | 10,632 | 10,632 | 0 |
| bars_unread | 23 | 23 | 0 |
| notes | 98,688 | **98,704** | **+16** |
| beats | 82,764 | 82,780 | +16 |
| bars_overfull | 1,569 | 1,569 | **0** |
| bars_short | 4,219 | **4,207** | **-12** |
| bars_defective | 5,344 | **5,332** | **-12** |
| bars_padded | 3,589 | 3,577 | -12 |
| inferred_rest_quarters | 4,885.0 | 4,877.0 | -8.0 |
| `unison_digits_shared` | 0 | **16** | +16 |

**Exactly two scores of the 293 change**, compared row by row from the per-score dumps rather than inferred from the totals:

- **Score A** — the table above.
- **Score B** — `notes` 387 → 391, `beats` 302 → 306, `inferred_rest_quarters` 20.75 → 18.75. **Every conformance figure of its own unchanged**: bars 42, defective 17, overfull 4, short 15, padded 14. Four notes come back into a voice that had been padded around them.

The +16 was **not** predicted from Score A alone (+12), so the extra 4 were traced to their score and their onsets before being accepted. `unison_digits_shared` equals the note delta exactly (12 + 4 = 16), which is the standing check that no note came back by any other route.

### 3. #116's pinned behaviour — all green, run explicitly

| guard | before | after |
|---|---|---|
| `unison_voices` fixture (8 bars, 32 coincident pairs) | 96 notes / 96 beats / 0 defective | **identical** |
| `test_a_unison_shared_by_two_voices_splits_across_their_stems` | pass | pass |
| `test_a_unison_shared_by_two_voices_is_overfull_without_the_stem_split` | pass | pass |
| Born a Stranger | 39 bars / 38 defective / 403 notes | **identical** |
| Carulli-Moderato-Op192 | 42 bars / 0 defective / 308 notes | **identical** |
| Score C (`coincident_unsplit_pairs` 15 / 4 staves) | pass | pass |
| Spanish Romance | 32 bars / 312 notes / 1 defective | **identical** (now pinned) |

Carulli had to be earned rather than inherited: the unguarded first version moved it to 382 notes / 2 defective.

### 4. New engraved fixture: `unison_in_chord`

`unison_voices` with the upper voice thickened into a chord, so the unison is one **member** of it. Engraved by the same MuseScore 4.6.3 the other fixtures name; only this one PDF engraved, every other fixture's bytes untouched.

| | unfixed | fixed |
|---|---|---|
| bars | 8 | 8 |
| notes | 96 | **128** |
| beats | 64 | 96 |
| bars_short / defective | **8 / 8** | **0 / 0** |
| `unison_digits_shared` | 0 | **32** |

Note what does *not* move: `bars` stays 8 and nothing goes overfull. The stem split resolves this page perfectly while a third of its notes are missing — which is why a fixture checking only stem binding would have passed.

### 5. Suite

- Baseline `16a33dd`, library set: **722 passed, 1 failed, 13 skipped**.
- This branch, library set: **733 passed, 1 failed, 13 skipped** (+11 tests).
- Without a library (CI's shape): **693 passed, 1 failed, 53 skipped** — two more library skips than baseline, the new Score A and Spanish Romance guards.
- The one failure is identical in all three and **pre-existing**: `test_the_summary_says_how_many_tests_skipped_for_want_of_a_library`, which fails on this machine only because `web/node_modules` is absent so the summary carries the alphaTab skip line. Unrelated; not touched.

## Known limitation, recorded

Sharing only reaches a coincident copy that sorts **last** in pitch order — the unison on a chord's **lowest** member — because the rank match consumes heads from the top. A unison on a **top or middle** member gets nothing: both copies sit inside the matched slice and take two different digits between them, and the starved head is a different notehead with no twin. Measured: **48 of the 65** in-chord onsets, 47 with the unison on top and 1 in the middle, 34 of them on Spanish Romance. Every one reads **byte-identically to main**, so it is pre-existing and not a regression — written into `_share_unison_digits`'s docstring, filed as **#141**, and pinned by a Spanish Romance guard so it cannot drift while open.

The full breakdown of the 65, corrected from an earlier draft that called 49 of them unsplit pairs:

| | n |
|---|---|
| twin in another voice — digit shared | **16** |
| no coincident twin at the leftover head's position (#141) | **48** |
| both copies in one voice, an unsplit pair — refused | **1** (Kids Run Through the City Corner) |

## Mutation testing

Each applied to the source, run, then reverted and re-run green. The red sets **overlap** — every one of them includes the library-wide aggregate, which is what that test is for — but each has its own signature beyond it.

1. **Drop the sharing entirely** → 4 red: the engraved fixture, the decode-level sharing test, the Score A guard, the library aggregate.
2. **Drop the chord condition** (share on any coincident twin) → 4 red: **Carulli's pinned guard**, the pair-alone boundary test, **the Spanish Romance guard** (notes 312 → 367, defective bars 1 → 28, shared digits 0 → 55), the library aggregate — while the fixture and the Score A guard stay **green**. The guard is load-bearing and pinned in both directions.
3. **Drop the same-voice refusal** → 2 red: the unsplit-pair test, the library aggregate.
4. **Force `unison_digits_shared` to 0** → 3 red: the fixture, the Score A guard, the library aggregate.
5. **Remove `unison_digits_shared` from api.py's `_BAR_KEYS`** → 3 red, all in `test_transcription_api.py`: the reload path is genuinely exercised, not pinned only where the value is 0.

## Test plan

- [x] `unison_in_chord` fixture: 8/8 short and 96 notes unfixed, 0/8 and 128 notes fixed, 32 shared digits disclosed
- [x] Decode-level: an in-chord coincident copy takes its twin's digit; a pair alone at an onset does not; a pair that stayed in one voice does not; a chord the tab never fully named is not patched — each asserting the disclosure count too
- [x] Score A guard (78 bars / 914 notes / 23 defective / 19 short / 12 shared)
- [x] Spanish Romance guard for the #141 family (unchanged, 0 shared)
- [x] #116's `unison_voices` fixture, Born a Stranger, Carulli-Moderato, Score C — all unchanged
- [x] Library-wide aggregate re-pinned with both moving scores named, plus the shared-digit total
- [x] `unison_digits_shared` reload path through api.py `_BAR_KEYS`
- [x] Mutation: five reverts, each with its own named signature

Closes #137. Follow-ups: #140 (the whole note #137 also named), #141 (the arrangement this cannot reach).


sirged:
Adversarial review follow-ups are on the branch (5185dd9), plus the corrected PR body above.

**Material finding — accepted, and I reproduced it before changing anything.** `:16 3.1 :1 5.1` is present at `d07387d`, `ee43e1c`, `16a33dd` and this branch. My "does not reproduce" was a method error: I diffed split-against-nosplit, saw the whole note in neither side of the diff, and generalised that into a claim about the score. It is in both sides. Also confirmed your token accounting — 5 `:1` tokens at `d07387d`/`ee43e1c` of which 2 are whole rests, 3 whole notes; the 2 rests are #134's doing.

Traced it to bar 13 voice 1 (0.25 + 4.0 = 4.25 in 4/4, one of the 8 overfull bars) and, since the page check was the open question, rendered it: it is a **harmonic** — grace note with a slashed 16th flag into a hollow notehead on a ledger line, `5` on string 1 — whose hollow head resolves to `notehead_whole` (gid 84, no stem, no candidate stems) and is emitted at 4.0 quarters. That is the harmonic-notehead gap the fixtures README already names, now with a measured cost. Filed as **#140**; the PR body and the Cosmic Wheel guard's docstring both point at it, so #137 does not close with that symptom untracked.

**Nit 1** — corrected. Measured 48 no-twin / 16 shared / 1 same-voice (Kids Run Through the City Corner, FF VI), not 49 unsplit pairs. 500/65/16 were right and are unchanged.

**Nit 2** — `unison_digits_shared` added through `ExtractionResult`, `to_dict()` and api.py's `_BAR_KEYS` (and the mirrored `BAR_KEYS` in `test_transcription_api.py`, so the reload path is real). Measured 12 / 4 / 16, 0 on the other 291. Pinned on the fixture (32), the Cosmic Wheel guard, the library aggregate, and all four decode-level tests. Two new mutations: forcing it to 0 turns 3 red; removing it from `_BAR_KEYS` turns 3 API reload tests red.

**Nit 3** — written into the docstring with the measured breakdown (47 top-member, 1 middle, 34 on Spanish Romance) and filed as **#141**. The evidence did support it as a distinct defect, so it is an issue rather than only a comment — with the explicit caveat that the 48 need the same page check #116 did before any is called wrong. Pinned by a new Spanish Romance guard (32 bars / 312 notes / 1 defective / 0 shared) so it cannot drift while open.

**Nit 4** — both fixed in the body: the red sets are described as overlapping (all include the library aggregate) rather than disjoint, and the no-library line now states its pre-existing failure.

Suite with the library: **733 passed, 1 failed, 13 skipped** — the failure being the same pre-existing `node_modules` one as baseline. Not merging.

Issue #163, opened 2026-08-29
A rest_whole glyph is read at its face value even beside other events in its bar's voice


sirged:
Split out of #109, which names it but is about something else. #109's headline — seventeen dotted whole notes in bars read as 4/4 — turned out to be a meter misread and is closed by #90/#104 (see the PR for #113, which pins the self-check library-wide). This is the residual that issue describes in passing and that the pin still finds.

**Correction (adversarial review of #166, which filed this issue): the "alone in its bar's voice" premise below is false at both named sites.** Re-extracted in-worktree and printed the emitted voice contents directly rather than trusting the earlier description. Neither rest is alone; both share their voice with other events, and the proposed rule would never have fired on either one. The real sites and the real question are below.

## The sites, as actually emitted

**Classical-Guitar-Method-Vol1-2020**, bar 16, voice 1 (3/4):

```
eighth, eighth, eighth, WHOLE REST, eighth, eighth, eighth
```

0.5×3 + 4.0 + 0.5×3 = **7.0 quarters in a 3/4 bar**. The whole rest is not alone — it sits between two runs of three eighths in the same voice. (The three eighths on either side already sum to exactly 3.0 on their own, which is the bar's whole length — a detail worth keeping in mind for whoever picks this up.)

**A second library score**, bar 5, voice 1 (4/4):

```
DOTTED WHOLE REST, quarter, quarter, quarter
```

6.0 + 1.0×3 = **9.0 quarters in a 4/4 bar**. Same shape: the rest sits beside three quarter notes in its own voice, not alone.

## Why the original proposal is wrong

> a `rest_whole` glyph that is alone in its bar's voice takes the bar's length

This never fires here. Both rests share their voice with other written events, so "alone in its voice" is not the condition that distinguishes a legitimate whole-measure rest from an over-long one — at least not at these two sites, which are the only ones this library has.

## The real question

[...]

Issue #162, opened 2026-08-29
A courtesy meter is applied a system early, and two whole notes overflow their bar on a library score


sirged:
Found while pinning #109's self-check — "a note written longer than the bar it sits in" — over the whole library. The seventeen dotted whole notes that issue counted are gone (they were a meter misread, closed by #90/#104; see the PR for #113). Two sites of the same *class* remain, and this is one of them.

## The site

**A library score**, bars 43 and 44, both read as **3/4** and both holding a **whole note** — four quarters in a three-quarter bar, which is impossible whatever the page says.

The page says the bars are in **4/4**, and that the decoder read them correctly otherwise: each is a whole-note bass pedal (D2, string 6 fret 0) under a run of eighths, and the eighths add to exactly 3.0, so with the whole note the voice holds 7.0 against a bar it thinks is 3.0.

## What is actually wrong

The meter. That score's *only* 3/4 is a **courtesy time signature** — the meter for the system that follows, engraved as the last thing on the system that precedes it, ahead of the `Coda` bar. It is being applied from bar 43, one bar into that last system, instead of at the Coda where it takes effect.

So this is not a duration defect and not a dot defect. It is the mid-system meter reader starting a change too early, which is exactly the hazard `server/tests/conftest.py` already names on the `kaine_salvation` fixture:

> A courtesy time signature - the key and meter for the NEXT system, printed as the last thing on THIS one... Widening the mid-system window enough to read a key change at a barline also brings this within reach, and reading it there would start the change a system early

Same shape, different score, and here it is not merely a risk — it is happening.

## Why it is worth its own issue

- It is **cheap to detect and cheap to verify**, like the rest of #109's class: the arithmetic proves the reading wrong with no reference to a page, and the page then says which half is wrong.
- The fix belongs to the meter reader (#90 / #104), not to the rhythm decode, so it should not be smuggled into a duration change. `server/tests/test_tabextract.py::test_no_emitted_note_is_longer_than_the_bar_it_sits_in` names these two bars as a known exclusion; closing this issue means deleting them from that list.
- The guard that protects `kaine_salvation` evidently does not generalise. Establishing *why* it holds there and not here is most of the work.

## Verifying a fix

- The library score's bars 43-45 read as 4/4, and its bar 46 (`Coda`) still reads as 3/4.
- `test_no_emitted_note_is_longer_than_the_bar_it_sits_in` passes with those two entries removed from its expected list.
[...]

PR #166, opened 2026-08-29
Count every stroke of a beam group, and ask what the output can never legally contain (#113, #109)

### LINKED ISSUE #109: Seventeen dotted whole notes are impossible in the bar they sit in
Found by the same technique that found #88 and #89 — asking what the output can never legally contain, then counting.

The library emits **52 dotted whole notes**. A dotted whole is six quarter notes long.

- **35** are in 12/8, where six quarters is exactly one bar. Correct, and a normal way to write a bar of 12/8.
- **17 are in bars read as 4/4**, where six quarters cannot fit.

So for those seventeen, either the meter is wrong or the dot is wrong, and the two candidates point at different existing issues: a meter misread is #90's territory (18% of first pages lose their printed meter and fall back to an assumed 4/4), while a spurious dot is #89's (augmentation dots binding to the wrong notehead in a chord).

Worth separating them before fixing either, because the answer tells you which. A 12/8 piece misread as 4/4 would produce exactly this shape and the dot would be *right*; a genuine 4/4 bar with a stray dot would produce it too and the meter would be right.

## Why it is worth its own issue

It is a self-check the pipeline can run on itself. Unlike most extraction defects, this one needs no reference to the printed page — the arithmetic alone proves something is wrong, which makes it cheap to detect and cheap to verify a fix against.

The same class is worth sweeping for more broadly. Two already-known members:

- **A rest longer than its bar.** #88 took this from 7 to 1. The survivor is a whole rest alone in a 3/4 bar, which is legitimate engraving — a whole-measure rest is drawn as a whole rest in any meter — and this decoder reads it as four quarters. That is a real remaining defect and arguably belongs here too.
- **A note longer than its bar.** 326 of them, which is too many to be one cause.

## Suggested shape for a fix

Whatever the cause turns out to be, the honest interim behaviour is the one this project keeps arriving at: where the arithmetic proves a reading impossible, say so rather than emitting it silently. An impossible duration in a bar is at least as good a signal as a bar that does not add up, and the machinery to report it already exists.

### LINKED ISSUE #113: Two adjacent unbeamed 32nd notes read as 16ths, doubling their duration
Found while attacking the flag rules on #108. Identical in that branch and in `main` — pre-existing.

An engraved fixture with two adjacent unbeamed 32nd notes decodes as `flags=2` where it should be 3. Three flags is a 32nd, two is a 16th, so each note reads twice its written length — a 2x error that overfills the bar and displaces everything after it.

Single unbeamed 32nds elsewhere in the same fixture read correctly, so the trigger is specifically **adjacency**: the flags of the neighbouring stem fall close enough to be miscounted or merged.

Unbeamed 32nds are rare in this library, which is why bar arithmetic has not been surfacing this — but they do occur in the classical studies, and a 2x duration error is the most damaging kind because it is silent and internally consistent.

### Verifying a fix

The fixture reads `flags=3` on both notes. Library-wide flag counts should be unchanged, since the fix is to a case that currently produces a wrong count rather than no count.

sirged:
Closes #113. Closes #109.

Two issues from the duration-misread cluster. One is a real, library-wide 2x duration error with a fix; the other stopped reproducing three months of commits ago and gets a permanent self-check instead of a repair. Both were reproduced against `origin/main` before anything was changed.

**This body has been corrected by an adversarial review pass after the original commits landed** (three factual errors found and fixed, plus several cheap secondary corrections). Everywhere a correction applies, the original claim is struck through or marked, not silently replaced, because the commit history under this PR still carries the uncorrected versions and that is what a bisect reads.

---

## #113 — a beam group's third stroke was never counted

**Reproduced, and the issue's stated mechanism is not the one.**

The issue reads: *"An engraved fixture with two adjacent unbeamed 32nd notes decodes as `flags=2` where it should be 3... the trigger is specifically adjacency: the flags of the neighbouring stem fall close enough to be miscounted or merged."*

The symptom is exactly right. The mechanism is not. Engraved both shapes and measured them:

| shape | decode at `origin/main` |
| --- | --- |
| two adjacent 32nds that MuseScore **beams** | `:16 :16` — **`flags=2`, wrong** |
| two adjacent 32nds that MuseScore **flags** (adjacent on the page, different beam groups) | `:32 :32` — correct |

Adjacency is not what does it. A 32nd's flag is a single glyph worth three hooks (`FLAG_HOOKS`) found by ink overlap with the stem's tip, and a neighbouring stem does not disturb it. What was wrong was the **beam** count — and MuseScore beams two adjacent 32nds, which is very likely what the fixture behind the issue actually held.

### The defect

A beam group is not one stroke at the tip of a stem; it is a **stack** that starts there and grows toward the notehead at a constant pitch, so the *n*th stroke sits *(n-1)* pitches in. `_beam_count_near` accepted a stroke only if it sat within `beam_y_tol` of the tip.

Measured over the whole library — 77,047 (stem, covering beam) pairs, offsets taken inward from the stem's tip, in staff spaces:

| band | pairs | peak |
| --- | --- | --- |
| at the tip | 55,391 | 0.0 |
| one pitch in | 12,267 | 0.75 |
| **two pitches in** | **366** | **1.5** |
| a different voice's beam crossing this stem | 8,881 | > 3.4 |

`beam_y_tol` is **1.17** spaces. It covers the first two bands and stops just short of the third. ~~So every three-stroke beam in the library lost a level~~ **Correction: not literally every one.** 42 stems already read 3+ levels via a different, pre-existing over-count (a neighbouring voice's beam passing beyond this stem's own tip), so the true miss is smaller than "every" claimed. What is unchanged: every note under a beam this window under-counted — a 32nd — was emitted as a **16th, at twice its written length**. Silent, internally consistent, and the most damaging kind of duration error, exactly as the issue says.

### The fix, and why not just widen the window

Widening is the wrong repair: the window is what keeps a neighbouring voice's beam out, and 8,881 pairs sit further in than 3.4 spaces. What identifies the third stroke is not its distance from the tip but that it **continues the run already accepted**. So that is what is asked. After the window has had its say, the count follows the stack inward while each next level is one pitch beyond the last.

This stack-following step is **strictly additive** — nothing the window accepts today can be rejected by it — so it can only add levels, which the library confirms: **366 stems across 49 scores gain exactly one level, and none loses any.** (A tip-anchored *replacement* rule was measured too; it also gains the 366, but it takes a level away from 105 stems, so it was not used.)

**Correction (adversarial review, "B1"): the additive claim above is true only of the stack-following step — it was never true of the tip-window check itself, and the tip-window check had its own bug.** The window comparison rounded the offset to 0.1pt *before* comparing it against `beam_y_tol`, rather than after — rounding units are points, not staff spaces — so a genuine offset could sit inside the tolerance while its rounded form sat just outside it, and get rejected outright. Measured: one library score, at this library's most common staff spacing (5.125pt), has a stroke at offset −5.9711pt against a tolerance of 5.99625pt — inside the window unrounded, but rounding to −6.0pt put it outside. That cost the stem its *only* level (an eighth would have read as a quarter; output-neutral on this particular score only because emission happened not to change there). Fixed by comparing the unrounded offset and rounding only the kept value — verified to keep the 902/0/0 suite and every library pin green, plus a new boundary unit test pinning exactly this case. **Note for anyone bisecting: commit `54d2823`'s own message states "Nothing the window accepts today can be rejected by that... none loses any" as an unconditional claim. That statement is the one this correction narrows — trust this PR body and the current code comment over that commit message.**

`beam_stack_pitch` sits on a wide plateau: ~~every value from 0.8 to 2.0 staff spaces counts the identical set~~ **Correction: measured more precisely, it is [0.76, 3.5] staff spaces**, because nothing a stem could wrongly reach lies within 3.5 spaces of its own deepest stroke; below 0.76 nothing is added at all. Above 3.5 a different failure mode appears before the library's own false positives would: at pitch 4.0, a genuine *two*-level beam can chain twice into a spurious *four* (verified directly against `_beam_count_near` — the stack-follower cannot yet tell "one more real stroke" from "an unrelated stroke that happens to sit within the same oversized window"). It is set to 1.10, inside the true plateau with margin on both sides.

### Library figures — full 297-PDF extraction, before and after

| | before | after |
| --- | --- | --- |
| `bars_overfull` | 1600 | **1541** |
| `bars_defective` | 5340 | **5300** |
| `bars_short` | 4188 | 4190 |
| `bars_padded` | 3602 | 3603 |
| `inferred_rest_quarters` | 4892.125 | 4897.875 |
| `notes` | 99461 | 99461 |
| `beats` | 83365 | 83365 |

A census of every emitted `<note>` by written value says the same thing more sharply. Across all 293 extractable scores, **exactly two entries change**:

| written value | before | after |
| --- | --- | --- |
| 16th | 12718 | **12373** (−345) |
| 32nd | 1590 | **1935** (+345) |

Nothing else moves at all — no other value, no dotted variant, no rest. 345 notes stop being emitted at twice their length, and that is the entire effect on the output.

`notes` and `beats` do not move, and that is the check that this corrects durations and nothing else: `beats` counts **onsets** (`musicxml.written_beats`), and halving a note's length neither adds nor removes one. `bars` is unmoved. Every counter that is not a Rule 8 conformance figure — dots (9182), `dots_unassigned` (1154), repeats (533), endings (492), segno (88), coda (150), navigation, `unison_digits_shared` (16), `systems_unread` (2) — is identical, compared **score by score** with per-score hashes: 244 of the 293 extractable scores come out byte-identical, and all 49 that differ differ only in note durations.

### Page checks — 12 changed bars, 6 scores, counted stroke by stroke

Score J 7/8/10, Troian Beauty 13, Stables 9/27, Score K 1, Serenade of Respite - Night 9/27, Singing of the Gentle Wind 2/11/18 — read off the rendered pages and counted against the vector beam geometry rather than by eye.

**All 12 are engraved with three beam strokes**, so the notes this change re-read really are 32nds in every case. ~~10 of the 12 now match the print exactly and sum to their meter.~~ **Correction: it is 8 of the 12, not 10.** Score R bar 1 and Singing of the Gentle Wind bar 18 each also keep a short voice 2 that this change does not touch, beside the two compensating-error bars already named below.

The other 4 are worth naming individually:

- *Serenade of Respite - Night* bar 9 and *Singing of the Gentle Wind* bar 2 each also **drop** a note (a second beam group collapsing into one), so the emitted bar still does not match the page. In Serenade bar 9 the dropped note cost exactly what the over-valued 32nds added — two notes moving from a 16th to a 32nd is −0.25, against a 0.25 shortfall elsewhere in the bar — so that bar used to add up **by coincidence**. That is a compensating error, not a regression: the durations underneath it are more right now than they were, and the note-drop it was hiding is a separate defect this change only stops masking rather than causes.
- Score R bar 1 and Singing of the Gentle Wind bar 18 each keep a short voice 2, unrelated to the beam-stack fix.

~~The other 2 are worth naming, and are what the `+2 bars_short` is.~~ **Correction ("B2"): Serenade bar 9 and Singing bar 2 are NOT what the +2 in `bars_short` is.** Both bars were already short before this change, short in a *different* voice than the one this fixes (Serenade voice 2 sits at 2.5 of 3.0; Singing voice 1 at 2.75 of 4.0), so reading their 32nds correctly does not flip either bar's already-short verdict — their net contribution to `bars_short` is zero. The two sites that actually flip a bar to short: **Midnight Harmony (New World) bar 13** (voice 1: 4.125 → 3.75, now short where it used to overfill) and **Singing of the Gentle Wind bars 8 and 26 together** (net +1 across the two). Midnight Harmony is not named anywhere else in this PR or its commits — it is named here because this is the only place its own part of the fix shows up.

Honest hit rate: **12/12** on "these notes are 32nds"; ~~10/12~~ **8/12** on "and the emitted bar now matches the print".

### Known residuals (not fixed here)

The Girl Who Stole the Stars, p1, x=430.0 and x=442.2: both print three beam strokes and this PR improves their read from 1 level to 2, but neither reaches 3. Not investigated further as part of this change.

### The fixture

`thirty_second_beams` is the first fixture in this repository to engrave a three-stroke beam — `rests_and_flags` separates every 32nd from the next with a rest, so no beam was ever drawn over one. It carries the **flagged** case beside the **beamed** one on purpose: bar 4's two 32nds are adjacent on the page but in different beam groups, and they read correctly before *and* after. That pairing is what makes the diagnosis decidable, and a fixture with only the beamed half would have left "adjacent 32nds are misread" standing.

At `origin/main` it reports 6 overfull bars and reads every beamed 32nd as a 16th; with the fix all 8 bars read back exactly as engraved, 0/0/0.

Two smaller things it needed, both measured rather than guessed:
- **16 divisions per quarter.** The generator's 4 rounds a 32nd's duration to **zero** — which is what the committed `rests_and_flags.musicxml` writes for each of its. Scoped to this fixture with a context manager, so every other committed fixture's MusicXML stays byte-identical (`--check` confirms).
- **One bar per system.** At four bars to a system MuseScore packs consecutive 32nds close enough that the decoder reads a beamed pair as one chord and drops the note after it — a coincident-onset defect landing on top of the one under test. The alternatives were engraved and read back, not judged by eye.

A four-stroke beam is deliberately *not* in the fixture: the rule counts one correctly (there is a unit test on constructed geometry), but the emitter's whole duration vocabulary stops at a 32nd (`musicxml.TYPE_NAMES`), so a 64th would be read as four levels and emitted as a 32nd regardless. Recorded in the fixtures' known-gaps list rather than asserted — **and, per the same adversarial review, the fixtures README's known-gaps entry was found overclaiming this: no genuine 64th exists anywhere in the library to confirm that against. The only stems this decoder reads at four levels are Troian Beauty p2's grace notes, a pre-existing, unrelated over-count. Corrected in the README.**

---

## #109 — the seventeen no longer reproduce

**This PR does not fix #109. It establishes that #90/#104 already did, and leaves behind the self-check the issue asked for.**

#109 counted 52 dotted whole notes, 35 legal in 12/8 and **17 in bars read as 4/4**, and named two candidate causes: a misread meter (#90) or a stray augmentation dot (#89).

It was the meter. Measured by extracting the same 297-PDF library against `git archive` of each tree:

| tree | dotted wholes in 12/8 | **impossible** |
| --- | --- | --- |
| `b13ffd1` — the commit #109 was filed against | 28 (+15 double-dotted) | **17** |
| `8101b00` — #90/#104, the printed meter, two days later | ~~56~~ **41 (+4 in 6/4)** | **0** |
| `21a1170` — current `main` | 56 (+8 in 6/4) | **0** |

**Correction: the `8101b00` row above previously repeated the `21a1170` row's figure (56).** Re-extracted directly against `git archive` of `8101b00`: 41 dotted wholes in 12/8, plus 4 in 6/4 — 45 total, not 56. The `21a1170` row's own 56 (+8 in 6/4) figure was already correct and is unchanged; only the earlier commit's row was wrong.

The seventeen are exactly reproducible at the commit the issue was filed against — five on *Score O*, eight on *Score P*, four on *Score Q* — and gone at the one that landed the printed-meter reading. **All three scores were read off their pages:** *Score P* and *Score O* are engraved in **12/8** and were being barred as 4/4; *Score Q* prints **6/4** meter changes. Six quarters is exactly one bar in all three, so every one of the seventeen is a dotted whole that exactly fills the bar it is printed in.

A neighbouring impossibility the issue did not count went the same way: at that commit fifteen **double**-dotted wholes — seven quarters — sat in 12/8 bars, and #111/#112's dot binding took those to zero too.

### What lands instead

`test_no_emitted_note_is_longer_than_the_bar_it_sits_in` — the self-check #109 itself proposed (*"a self-check the pipeline can run on itself... the arithmetic alone proves something is wrong"*), run over the whole library, written over **every** written value rather than dotted wholes alone. It goes red against the pre-#90 decoder.

It found three residual sites, all named in the test rather than assumed away by a looser rule, and all given homes:

- **Score L bars 43-44** — a whole note in a bar read as 3/4. The page says 4/4 and says the decoder read the bars correctly; what is wrong is that the score's only 3/4 is a **courtesy meter** for the Coda, being applied a system early. That is the hazard `conftest`'s `kaine_salvation` fixture already names, failing on a different score. Filed as **#162** — it belongs to the meter reader, not to any duration.
- **Classical-Guitar-Method-Vol1-2020 bar 16** — a whole rest, read as four quarters in a 3/4 bar. ~~Alone in its bar's voice~~ **Correction ("B3"): not alone.** Re-extracted directly and printed the emitted voice: it reads `eighth, eighth, eighth, WHOLE REST, eighth, eighth, eighth` — the rest sits between two runs of three eighths in the same voice, which already sum to the bar's full 3.0 quarters on their own. Filed as **#163**, whose own proposed rule ("a `rest_whole` alone in its voice takes the bar's length") has now been corrected there as well, since it would never have fired at this site. Deliberately not fixed here, because it would move the library's conformance figures a second time in one change and leave neither movement separable from the other.
- **Score M bar 5** — a *dotted* whole rest, six quarters in 4/4. Also **not alone**: re-extracted, it reads `DOTTED WHOLE REST, quarter, quarter, quarter` in its voice — 9.0 quarters total. Carried on the corrected #163 beside the one above.

---

## Verification

- **Suite:** 903 passed, 0 failed, 0 skipped, with `FERMATA_TEST_LIBRARY`, `FERMATA_MUSICXML_XSD` and `web/node_modules` all present. Baseline before this change was 893/0/0; the original commits added 9 (five new tests, one new fixture entering four parametrized lists), and this adversarial-review pass adds one more — a unit test pinning B1's tolerance-rounding boundary.
- **Mutation, commit-backed, re-run after this review pass.** Reverting `_beam_count_near` entirely back to its pre-#113 tip-only form, with everything else in place: `test_a_thirty_second_under_a_beam_is_not_read_as_a_sixteenth` red, `test_a_beam_group_is_counted_to_its_full_depth` red, `test_library_wide_repeat_structure_leaves_conformance_untouched` (the library conformance pins) red (`bars_overfull` reverts to 1600) — while `test_the_flagged_thirty_seconds_beside_them_were_never_wrong` stays **green**, which is the control doing its job, and #109's self-check stays green. Restored, rebuilt, green (903/0/0).
- **B1's own mutation.** Re-applying round-before-compare to the tip-window check alone (leaving the stack-following logic untouched): the new boundary unit test goes red; restored, green.
- **#159's dot machinery** re-run explicitly and green: Storm's Past, Score N, the guard-5 saves, and the `dots` / `dots_unassigned` / `dots_unassigned_no_candidate` / `dots_unassigned_eliminated` library pins, none of which this change touches.
- **Stale values, re-grepped after this review pass.** `docs/musicxml-tab-profile.md`'s re-measured round-trip note quoted `bars_defective` at 5,340 and now says 5,300, with the cause named in the same sentence as the others — unaffected by this pass, since no pinned library total moved, only attribution prose. The `8101b00` table row above and the fixtures README's 64th known-gap entry were the two stale copies this pass found and fixed; repo-wide grep for both figures' old values and for "10 of the 12" / "10/12" turned up no other copies.


sirged:
Do not merge yet — Browser tests were still running at the time of writing. Every other check passed, and the server-test collection matches `main` exactly:

| | passed | skipped | collected |
| --- | --- | --- | --- |
| `main` (21a1170) CI | 835 | 58 | 893 |
| this branch CI | 843 | 59 | **902** |

+9 collected: eight new tests that run in CI, and exactly one (the library-gated self-check) that skips there, which is the intended shape. Locally, with `FERMATA_TEST_LIBRARY`, `FERMATA_MUSICXML_XSD` and `web/node_modules` all present, the run is **902 passed, 0 failed, 0 skipped**.

Issue #195, opened 2026-08-31
Chord flash cards: musical-style groupings, and barre/open sevenths beyond dominant


sirged:
Follow-up to #28. The chord flash card drill ships three family presets - major & minor, (dominant) sevenths, and barre chords (major/minor, E-shape and A-shape, generated across the fretboard) - which covers the issue's "majors and minors first, then sevenths, then barre chords" ordering.

Two pieces of #28's own scope were left for later, deliberately:

- **By musical style** (blues, bossa nova, folk, ...), which the issue itself frames as an optional grouping. Doing this honestly needs a real per-chord style tag on top of `chord-shapes.js`'s library, not a guess bolted onto the existing family presets.
- **Barre minor7/major7, and open minor7/major7 shapes.** Today's sevenths family is dominant-7 only (the open E7/A7/D7/G7/B7 shapes), and barre shapes are major/minor only. `chord-theory.js`'s `QUALITIES` already has the interval math structure to add `minor7`/`major7`; what's missing is real fingerings for them (open and barre) checked against that math the same way `chord-shapes.spec.js` checks the current library.

Neither blocks the drill working today - see #28's PR - but both round out the "chord families" and "musical style" asks in the original issue.

Issue #194, opened 2026-08-31
Named, saved constraint presets reusable across drills


sirged:
Follow-up to #26. The shared constraint model (string set, fret range, and a key/note-set filter) now exists in `web/src/lib/trainer/constraints.js` and is used by both the fret-to-note drill and the chord flash cards - a scope object with the same shape narrows either one.

What's still missing is the other half of #26's ask: **naming a scope and saving it**, so a player can build "open position, key of G" once and pick it again in a different drill later, rather than re-selecting strings, a fret range and a key by hand every time.

Needs, roughly:
- A place to store a named preset (owner-scoped, like everything else here until real accounts exist) - a small table alongside `instruments`/`practice_goals`, or an extension of settings.
- An API to list, save, and delete presets.
- A picker in each drill's scope controls: choose a saved preset, or build one by hand and save it.

Not attempted here because it's a distinct, separable piece of work (storage + API + UI for naming and reuse) from the constraint arithmetic itself, and the two drills that exist today already prove the arithmetic is right without it.

Issue #146, opened 2026-08-28
The one disclosure #139 added reaches the reader tuples but never the stored row


sirged:
Found during the #145 review, confirmed empirically through the production route.

**Symptom:** `unison_digits_shared` (added in #139, merged as 560f030) stores and reads back as `None` in production for every score — including where the decoder inferred digits (one score: direct extraction says 12, POST and GET both return `None`, the key is absent from the stored blob). The warning prose IS emitted, so the disclosure is half-delivered: the sentence reaches a reader, the structured count reaches nobody. Third recurrence of exactly the failure mode `api.py`'s own comments name.

**Cause:** the field is on `ExtractionResult`, `to_dict()` and `_BAR_KEYS`, but absent from the hand-written `confidence_json` dict in `api.transcribe()` — the only write path into storage. `to_dict()` is never called in `server/`, so presence there carries no production weight.

**Scope:** exactly one field. Verified exhaustively by AST-diffing the `confidence_json` literal against `to_dict()`: no other key is read back but unwritten (after #145, which fixed the same gap for #116's pair).

**Why three prior tests didn't catch it:** the `BAR_KEYS`-iterating reload tests assert key presence and self-equality — both sides of their comparison come from `_transcription_dict` over the same stored row, so an unwritten field compares `None == None` and passes. #145's test shape (a literal, independently derived value asserted against a post-storage read) is what catches this class.

## Fix

One line in `confidence_json`: `"unison_digits_shared": result.unison_digits_shared,`

## Regression test — can be ungated

`server/tests/fixtures/engraved/unison_in_chord.pdf` is a committed fixture yielding `unison_digits_shared == 32` — no `FERMATA_TEST_LIBRARY` needed. Transcribe it through `api.transcribe` → `api.get_transcription` and assert the literal 32 on the post-storage read.

## Close the bug class, not just the instance

Add a structural guard: assert the set of keys `transcribe()` writes into `confidence_json` is a superset of `_BAR_KEYS`, so the next field added to the reader tuple cannot repeat this. This is the third disclosure counter in a row to hit some form of the gap (#134's fields, #116's pair via #143, now #137's counter) — the guard is the actual fix.

Issue #55, opened 2026-08-20
Transcribe in bulk instead of one score at a time


sirged:
Transcription is currently one score at a time, initiated by hand. On a real library that means nothing is ready to use: a freshly scanned 300-score library has zero transcriptions until someone clicks through all of it.

Bulk conversion, so a library becomes usable without per-score effort:

- Convert everything eligible in one action, and convert newly ingested scores as part of bringing them in.
- Run as a background job with progress and a running count, following the pattern the library scan already uses rather than adding a queue.
- Skip what cannot be converted, and skip what already has a transcription unless a reconversion is asked for.
- **Never overwrite an edited transcription.** The stored format already separates extracted from edited and a test guards it; bulk conversion must respect that or it will destroy corrections at scale, which is far worse than doing nothing.
- Report the outcome in aggregate: how many converted, how many were skipped and why, and how many came out with bars that do not add up, so the state of the library is visible without opening scores one by one.

Reconversion needs to stay cheap and safe, because the extractor will keep improving and a library should be able to benefit without losing work.

PR #196, opened 2026-08-31
Scope any drill by string, fret and key, and add chord flash cards

### LINKED ISSUE #28: Exercise: guitar chord flash cards
Flash cards for guitar chord shapes, in both directions: name the chord from a shape, or produce the shape from a name.

- Show the fingering on the neck when revealing, so the shape is learned visually rather than as a word.
- Hear it, using the synthesiser already present, since a chord recognised by ear and by shape is better known than one recognised only by sight.

## Scoping, which is what makes it useful rather than a quiz

- **A region of the neck** — open position, or up around the fifth, so shapes are learned where they will be played.
- **Chord families** — majors and minors first, then sevenths, then barre chords, so difficulty is chosen rather than random.
- **By musical style**, as an optional grouping: the shapes that turn up constantly in blues are not the ones that turn up in bossa nova or in folk. Practising a set that belongs together is more motivating than a shuffled deck, and it is closer to how people actually learn chords.

Right and wrong marked immediately, accuracy tracked, and the shared constraint model doing the scoping so a preset carries across exercises.

Depends on the interactive neck component and the constraint model.

sirged:
Closes #28. Progresses #26 - see below for what's left of it, filed as #194.

## What this does

**#26 - constraint model.** Factors fret-to-note's existing string-set/fret-range scoping out of `fret-to-note.js` into a reusable module, `web/src/lib/trainer/constraints.js`, and widens it with a key/note-set filter (major or minor key, any of the 12 roots) that neither drill exposed before. `fret-to-note.js` now imports `scopePositions`/`scopeIsAskable`/`scopeLabel` from there and re-exports them under the same names, so it is a pure move - nothing about the existing drill's behaviour changed. `groupInScope` is the one new primitive this needed that a single-position drill never did: whether every position in a *group* (a chord's whole shape, not one tap) fits inside a scope at once.

Still open: **named, saved presets** ("build this once, reuse it in a different drill later") - the constraint *arithmetic* is shared and tested, but there is no storage/API/UI yet for naming and reusing one. Filed as #194 rather than guessed at here.

**#28 - chord flash cards.** Built on the neck (#25/#184) and the constraint model above:

- `chord-theory.js` - pure interval arithmetic: `chordTones(root, quality)` for major/minor/dominant7, mirrored character-for-character in `server/fermata/trainer.py`'s `chord_tones`.
- `chord-shapes.js` - a real fingering library: hardcoded common open chords (E/A/D/G/C major, E/A/D minor, E/A/D/G/B7), plus **generated** movable E-shape/A-shape barre chords (major/minor) at every fret a region allows - the barre shapes' roots are read straight off the neck's own tuning arithmetic rather than a second table, so they are correct by construction, and every shape (open and barre) is checked against `chordTones` in tests.
- `chords.js` - the drill: three family presets (major & minor / sevenths / barre chords, matching the issue's own ordering), scoped by region + key via `constraints.js`, both directions (`shape_to_name`, `name_to_shape`), one grading rule for both - tone-set equality, never one canonical fingering.
- `ChordFlashCards.svelte` - reuses `Neck.svelte`'s `markers` prop for both a shown shape and accumulated taps; "Hear it" plays the chord via a new `playChord` in `score-render.js` (the same one-shot-MIDI pattern as the existing `playPitch`, widened to several simultaneous notes).
- Structured results: a new `trainer_chord_attempts` table (`server/fermata/db.py`), sibling to `trainer_attempts` rather than forced into its single-note columns - see that table's own comment for why, and `docs/practice-data.md`'s new section. `POST`/`GET /api/trainer/chord-attempts`, mirroring the fret-to-note routes. `correct` is always computed server-side from tone sets, never trusted from the request.
- `activity: 'chords'` for the practice session (already in `docs/practice-data.md`'s vocabulary, unused until now).
- No streaks/bests/averages/shaming - `chords.spec.js` and `chord-flashcards.spec.js` extend the FORBIDDEN_WORDS check from #184 to this drill's own wording.

Filed #195 for what is left of #28's own scope: musical-style groupings (which the issue itself calls optional) and barre/open minor7/major7 shapes (today's sevenths are dominant-7 only).

## Per-target

**Chord correctness (exhaustive, both languages)** - `web/tests/unit/chord-theory.spec.js` and `server/tests/test_chord_theory.py`:

```js
expect(chordTones("C", "major")).toEqual(["C", "E", "G"]);
expect(chordTones("A", "minor")).toEqual(["A", "C", "E"]);
expect(chordTones("G", "dominant7")).toEqual(["G", "B", "D", "F"]);
expect(chordTones("E", "major")).toEqual(["E", "Ab", "B"]); // Ab not G# - PITCH_CLASSES' own spelling
```

plus every one of the 36 (root, quality) combinations checked for note count, no repeated tone, starting on its own root, and the major/minor-triads-differ-only-in-the-third and dominant7-is-major-plus-one-note relationships holding at every root. `chord-shapes.spec.js` then checks every shape's *actual sounded notes* (via `neck.js`'s own tuning arithmetic) against `chordTones` - open shapes and every generated barre shape.

**Constrained drill only offers what's in scope** - `constraints.spec.js`, `chords.spec.js` (unit, narrowing fret range/string set/key), and `chord-flashcards.spec.js` (browser: narrowing to open position drops the G-major choice, which needs fret 3, and keeps E major, which fits).

**Full prompt to answer to feedback to next, both directions, correct and incorrect** - `chord-flashcards.spec.js`, against the real backend and build. `name_to_shape`'s correct/incorrect positions are found via the neck's own rendered `data-note` attributes plus an *independently reimplemented* chord formula in the test file itself (not imported from the app), so this would catch the app's own chord math breaking rather than just restating it.

**Chord correctness is server-authoritative** - `trainer.normalise_chord_attempt` computes `correct` from tone sets; `test_chord_trainer_api.py` proves the signature has no `correct` parameter to smuggle one through, and that grading is by tone SET (a tapped G major voiced in a different order/octave than the "reference" one still grades correct - `test_name_to_shape_grades_by_tone_set_not_by_exact_fingering`).

One honest caveat on that: for `name_to_shape`, the server compares the CLIENT's reported tapped notes (`given_notes`, resolved from the instrument's tuning on the client, the same trust boundary `trainer_attempts.given_note` already crosses for fret-to-note's note-to-position direction) against the target chord's own tones, computed server-side. The server has no instrument definition at grade time to re-derive those notes itself. What stays server-authoritative regardless is the verdict: `correct` is never a value the client can send, only the tone-set comparison this endpoint performs. For a self-directed practice drill (nobody but the player has a stake in the count), trusting the client's note-resolution while still computing the pass/fail server-side is the same shape of trust `checkTapAnswer`/`given_note` already established for the single-note drill - not a new gap this PR introduces.

**Empty/degenerate states honest** - a fret range or string set that admits nothing (`poolIsAskable` false, same "narrow-scope" messaging as fret-to-note); a key that shares no diatonic chord with a family preset (verified: F# major admits none of the open family's E/A/D/G/C roots).

**Structured results, field round-trips** - `test_chord_trainer_api.py`'s `test_logging_a_shape_to_name_attempt_stores_the_row_and_returns_it` and `test_logging_a_name_to_shape_attempt_round_trips_the_tapped_notes`: literal-value assertions on the POST response, a fresh SELECT against the row, and a subsequent GET - not just shape checks (#146).

**Mutations, each confirmed red then reverted via `git checkout` (commit-backed)**:
- Constraint filter (`groupInScope`'s `every` to `some`): `constraints.spec.js` and `chords.spec.js`'s scope tests go red.
- Chord-note math (major triad's third: interval 4 to 8, both JS `chord-theory.js` and Python `trainer.py`'s `CHORD_QUALITIES`): `chord-theory.spec.js`/`test_chord_theory.py` go red, and so does the *browser* suite (`chord-flashcards.spec.js`'s name_to_shape test, which computes tones independently as described above).
- Server grading (`correct = True` unconditionally): `test_chord_trainer_api.py` goes red.

**Suites vs. baselines.** Server: `PYTHONPATH=. py -3.13 -m pytest` gives 1124 passed, 64 skipped (pre-existing, library-gated, unchanged skip reasons) on the current rebase (main now carries #55's bulk-transcribe, +12 tests over the pre-rebase count). `test_every_route_has_exactly_the_expected_operation_count` re-derived against the live app on the actual rebase base (`app.openapi()['paths']`, counted independently of the test's own helper and cross-checked against it) - 60, not trusted from arithmetic: main's own count is 58 after #55's two routes, plus this PR's two (`POST`/`GET /api/trainer/chord-attempts`). Web: `npm run build` clean (no Svelte-warnings-as-errors); a full `npm run test:browser` run locally with a freshly rebuilt `dist`, isolated on a private port (never the suite's shared default 8931 or 8080) after discovering another concurrent worktree's leftover server squatting on 8931 earlier - not this branch's bug, an artifact of running several worktrees on one box. `#184`'s own fret-to-note/neck specs are untouched and pass. I'll add the final tail of this post-rebase local run as a comment once it finishes; CI runs the same suite independently regardless.

If you see one server-suite failure or an ~11-test skip-count delta locally: that's the pre-existing `conftest`'s `_library_skips` leak plus a `node_modules`-absence check, identical on `main`, and green in CI - not this PR's code.

## Rendered example

Could not attach a live screenshot from this environment (no connected browser session) - see the PR comments for a Playwright-driven render check and what it showed instead.

## Mandate

No AI-vendor/assistant wording anywhere in the new code, tests, or docs (grepped). No Co-Authored-By. `unset GH_TOKEN` before every `gh`/`git push`. Tested against 127.0.0.1, never `localhost`. Op-count re-derived live against the actual rebase base each time, not trusted from a comment or arithmetic. Spec floors added additively (`browser-chord-flashcards-28.json`, `unit-constraints-26.json`, `unit-chord-theory-28.json`, `unit-chord-shapes-28.json`, `unit-chords-28.json`), nothing existing lowered.

---

_Rebased onto current `main` (#55's bulk-transcribe now included) - op-count re-derived to 60, `server/fermata/api.py`'s import list and `test_api_docs.py`'s pin resolved additively, no other conflicts._

### DIFF
--- a/docs/practice-data.md
+++ b/docs/practice-data.md
@@ -427,16 +427,53 @@ right and wrong separately and never free to divide one by the other.
   this API keeping a bespoke aggregate endpoint in step with every future
   drill's own idea of what a weak spot is.
 
+## Trainer chord attempts
+
+The question this document's own "what is not here yet" section left open:
+whether a chord drill's attempts fit `trainer_attempts` or need a table of
+their own. *Chord flash cards* (issue #28) answered it - a chord does not
+fit. `trainer_attempts`' `target_note`/`given_note` are `NOT NULL` columns
+holding exactly one pitch class each, which is the right shape for a drill
+whose unit is one note; a chord is a SET of them, and forcing its tone set
+into a single-note column would either lose which chord was asked (keeping
+only its root) or store something that plainly is not a pitch class. So this
+is a sibling table, `trainer_chord_attempts`, built the same way
+`trainer_attempts` is: `correct` computed once, server-side, never trusted
+from a request body.
+
+| Column | Meaning |
+| --- | --- |
+| `id` | Stable for the life of the row. |
+| `owner` | `'local'`, same as everywhere else. |
+| `session_id` | The `practice_sessions` row logging the surrounding drill's TIME, when there is one yet - `NULL` otherwise, the same ordinary-not-edge case `trainer_attempts.session_id` is. |
+| `drill` | `'chord_flashcards'` today - a widened tuple, not a migration, is how a second chord-shaped drill would arrive, the same rule `trainer_attempts.drill` follows. |
+| `direction` | `shape_to_name` (shown a real fingering, asked to name the chord) or `name_to_shape` (named a chord, asked to place it on the neck). |
+| `target_root`, `target_quality` | The chord being tested, always set - a pitch class and one of `major`/`minor`/`dominant7`. |
+| `target_shape` | The fingering actually SHOWN - a JSON array of `{string, fret}` - set only on `shape_to_name`; `NULL` on `name_to_shape`, which shows no shape at all. |
+| `given_root`, `given_quality` | The chord chosen by name - set only on `shape_to_name`. |
+| `given_notes` | The pitch classes a TAPPED shape actually sounded, a JSON array worked out client-side from the instrument's own tuning - never from which shape the player meant to play, mirroring `given_note`'s rule above. Set only on `name_to_shape`. |
+| `given_shape` | The positions actually tapped, a JSON array of `{string, fret}` - set only on `name_to_shape`, kept beside `given_notes` so "which shapes get missed" stays a real question over this table. |
+| `correct` | Whether the given tone SET equals the target chord's own tones (`chord_tones(target_root, target_quality)`, mirrored character-for-character between `server/fermata/trainer.py` and `web/src/lib/trainer/chord-theory.js`) - the same rule either direction, computed here and only here. There is more than one right way to play a given chord, so this is never a check against one canonical fingering. |
+| `response_ms` | How long the question took, informational. |
+| `created_at` | When the row was written. |
+
+**One `correct` rule covers both directions, the same idea `trainer_attempts`
+already applies to a single note, widened to a set.** `shape_to_name`
+compares the CHOSEN chord's tones to the target's; `name_to_shape` compares
+what was actually TAPPED (resolved to pitch classes client-side, the same
+trust boundary `given_note` already crosses) to the target's. Tone-set
+equality, not label equality and not fingering equality, either way.
+
+### Asking questions
+
+- `POST /api/trainer/chord-attempts` - log one answered chord question.
+- `GET /api/trainer/chord-attempts?drill=&direction=&correct=&root=&quality=&session_id=&limit=` -
+  the raw, queryable record, newest first - the same shape
+  `GET /api/trainer/attempts` offers, with `root`/`quality` added so "which
+  chords am I weak on" is a filter rather than a client-side group-by.
+
 ## What is not here yet
 
-- **Interval or chord-recognition attempt shapes.** `trainer_attempts` is
-  drill-agnostic in its schema (a `drill` column, not a table per exercise),
-  but a chord drill's own idea of "which chord, which voicing" has not been
-  built against it yet, and inventing that shape before that trainer exists
-  would be the same guess this section spent a long time avoiding for
-  fret-to-note. Fret-to-note's `position_to_note`/`note_to_position` split is
-  a real answer for a single-note drill; a chord is a different unit, and #26
-  or #29 gets to decide whether it fits the same table or needs its own.
 - **Key, tempo and difficulty per score** - musical metadata about the piece
   rather than about the practice.
 - **Achievements** - looking back at what has been accomplished, where a goal
--- a/server/fermata/api.py
+++ b/server/fermata/api.py
@@ -65,6 +65,8 @@ from .api_models import (
     TagOut,
     TrainerAttemptListOut,
     TrainerAttemptOut,
+    TrainerChordAttemptListOut,
+    TrainerChordAttemptOut,
     TranscribeBatchStatusOut,
     TranscribeBatchTriggerOut,
     TranscribeResultOut,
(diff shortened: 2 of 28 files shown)


sirged:
Local full browser suite (unit + browser, tests/unit and tests/browser together, against a freshly rebuilt dist) finished clean on a private port:

    506 passed (5.4m)

Exit code 0 - the spec-floor guard (tests/spec-floor-guard.js) passed too, since it fails the run loudly on any shortfall and nothing printed. This includes every #184 fret-to-note/neck spec (untouched, still green) alongside the new constraint-model and chord-flash-card specs.

CI will run this suite independently on push; this is confirmation, not a substitute for that.

sirged:
Rendered example: could not attach an actual screenshot here - the claude-in-chrome browser extension is not connected in this environment, so there is no live browser session to capture from or an API path to upload an image to a comment. As a substitute, I served the built dist directly (127.0.0.1:8941, a throwaway scratch library, never 8080) and drove it with Playwright to confirm it actually renders:

- #/chords loads, "Shape → name" / "Major & minor" active by default, tuning/frets/strings/key controls all present.
- Clicking Start shows a real shape on the neck (open target markers, unlabelled - matching Neck.svelte's "never label a bare target" rule) and a grid of chord-name answer choices (A major, A minor, C major, D major, D minor, E major, E minor, G major - exactly the roots the open-shape library offers for this family).
- "♪ Hear it" is present and wired to the new playChord path.

This is the same rendering the 506-test local suite (see the earlier comment) already exercises end to end with real taps and real assertions on what's drawn - the screenshot was a visual sanity check on top of that, not a substitute for it.

sirged:
Post-rebase verification (onto main@496ef16, #55 included):

- Op-count re-derived live against the running app: **60** (main's 58 + this PR's 2 chord-attempt routes) - `test_every_route_has_exactly_the_expected_operation_count` passes.
- Server: `PYTHONPATH=. py -3.13 -m pytest` -> **1124 passed, 64 skipped** (skip reasons unchanged, pre-existing library-gated).
- Web: `npm run build` clean, full `npm run test:browser` (unit+browser, real build, private port 8942) -> **508 passed (5.6m)**, exit 0, spec-floor guard satisfied. `#184`'s fret-to-note/neck specs and everything else stayed green through the rebase.

No merge conflicts remain (`gh pr view` reports `mergeable: MERGEABLE`); `BLOCKED` is CI/review gating, not a conflict. Not merging, per instruction.

Issue #168, opened 2026-08-29
The side-by-side ArrowRight page-turn spec fails alone under CI load


sirged:
`tests/browser/practice-shortcuts.spec.js:677` — "side-by-side layout: PDF page-turning keeps its keys › ArrowRight turns the PDF page" — has now failed on CI twice in two days, each time alone, each time passing on re-run or in isolation, each time in a change that does not touch the PDF pane:

- 2026-08-28, PR #159's CI: failed once (5s timeout on the page HUD), re-run green. Flagged in that PR's report.
- 2026-08-29, PR #164's final CI (run 33243069393): `expect(locator).toHaveText(expected)` failed at 6.0s; 325 other tests passed. The PR touches `score-render.js` navigation, not the PDF pane.

This is the #110 pattern's remaining shape: the assertion retries, but whatever publishes the PDF page indicator under CI load sometimes takes longer than the retry window, or the keypress lands before the pane is ready to receive it. The spec needs its readiness barrier examined (what proves the PDF pane has focus/render before the ArrowRight is sent?), not a longer timeout.

## Verifying
- Root-cause the race (trace what the HUD text waits on); fix the barrier.
- The spec passes 20 consecutive runs under full-suite load (the reproduction condition both failures shared).

sirged:
Reopening: the race survives the #169 fix at low probability. Post-#169 sightings: PR #173's CI failed the same spec (practice-shortcuts.spec.js:677, and on one attempt its sibling at :707) on a base that includes the barrier fix — twice across two attempts — while main's own CI at the same base commits (af254ed run 33252165987, ecbfb76 run 33296002422) is green, and the branch passes locally 336/336 in full-suite runs. The failing branch touches no file on the PDF/ScoreCompare path (verified by diff), and its bundle grew <0.1%.

So the residual is load-marginal: #169's pendingPage fix closed the deterministic pre-render window (20/20 under forced delay), but something in the same path still loses the press occasionally under CI contention. The next investigation should start from a CI-shaped reproduction (the forced-delay harness from #169 at varying delays, plus CPU contention), and from the #169 review's note that the gig-mode test at :707 'shares the hole and has merely been winning the race' — both failures fit that description.

Issue #82, opened 2026-08-21
Browser test for the reference pitch bound fails at random


sirged:
`tests/browser/instruments.spec.js` — "a reference pitch outside the bounds is refused rather than silently defaulted" — failed once and passed on a plain re-run of the same commit, with no code change between the two:

```
TypeError: Cannot read properties of undefined (reading 'reference_pitch')
> const [saved] = await (await request.get("/api/instruments")).json();
  expect(saved.reference_pitch).toBe(415);
```

## Root cause — corrected

An earlier version of this issue blamed state leaking between tests. **That is wrong**, and the fixes it suggested would not have helped: there is already a `beforeEach` that deletes every existing definition, and the runner is pinned to a single worker precisely because instruments are global to an install.

The real cause is a missing synchronisation point. The test does:

```js
await page.locator(".actions button.primary").click();
const [saved] = await (await request.get("/api/instruments")).json();
```

`click()` resolves when the click is dispatched, not when the save it triggers has completed. The read then goes out of band — through the request context rather than the page — so it can and sometimes does overtake the POST. An empty list makes `saved` undefined, which is the error above.

The two neighbouring tests that read the API the same way do **not** flake, and comparing them shows why: each waits on the interface first, one for the instrument count to reach 1 and the other for the saved rows to render, before reading out of band. Those assertions are load-bearing as barriers, not just as checks.

## Fix

Put a barrier between the click and the read — waiting for the count attribute is enough and matches what the capo test already does. Waiting on the POST response would also work.

Worth fixing rather than re-running past: a browser job that fails at random trains everyone to re-run it, which is how a real failure gets waved through.

## Worth checking for elsewhere

Any test that performs a user action and then reads state through a channel other than the page is exposed to the same race. That pattern is what to grep for, rather than this one assertion.

Observed on https://github.com/sirged/fermata/actions/runs/32444208999 (first attempt failed, second attempt of the same run passed).

PR #198, opened 2026-08-31
Fix two #110 flakes, and add a guard against the pattern coming back

### LINKED ISSUE #110: A browser test can read state out of band without waiting for it, and the pattern keeps coming back
A test performs a user action, then reads the result through the request context rather than through the page. Clicking resolves when the click is *dispatched*, not when the request it triggers finishes, so the read is never ordered against the write and it sometimes overtakes it. The test fails at random, on a slower runner, in a way that looks like a product defect.

## Why this needs a mechanism rather than another sweep

Three instances, two sweeps, still recurring:

1. **#82** — the reference-pitch test. Diagnosed and fixed in #83, and the issue explicitly asked for the pattern to be grepped for: *"Any test that performs a user action and then reads state through a channel other than the page is exposed to the same race. That pattern is what to grep for, rather than this one assertion."*
2. **#100** — that grep was finally done and found exactly one remaining instance, the week-boundary goal test. Fixed.
3. **Now** — a refused-scan test, in a file written **after** that sweep. It waits for a client-side signal (an alert disappearing) and then reads the scan status out of band, which is not ordered against the server's rescan finishing.

So a point-in-time grep does not hold. The pattern is being reintroduced about as fast as it is removed, because the wrong version is the *natural* thing to write and the right version requires knowing this specific hazard.

## What a guard should do

Flag, at build or test time, a read through the request context that follows an action with no intervening wait on the page. The project already has this shape of check — `check-warning-patterns.mjs` fails the build when backend wording drifts out of step with the frontend's expectations — so a crude, greppable heuristic in the same style is consistent and cheap.

Being approximate is acceptable here. A false positive costs a moment and an explicit acknowledgement; a false negative costs a random CI failure that trains everyone to re-run rather than look, which is how a genuine failure eventually gets waved through.

Worth considering as the more durable alternative: a helper that makes the barrier the default, so the natural thing to write is the correct one. Every existing correct instance waits on the interface first — a count attribute, a rendered row, a tick — and the comments on those sites already say the assertions are barriers as much as checks. A named helper encoding that would carry the knowledge rather than relying on each author having it.

## Note for whoever takes it

The three fixed instances are worth reading first: they show the three different barriers that turned out to be correct in each case, and why a barrier has to be something that *cannot* be true until the server has answered. A wait on a client-side signal — an alert closing, a form resetting — is exactly the mistake the third instance made.

sirged:
Closes #110.

## The two named flakes, root-caused

**practice.spec.js:343** ("a finished week asks whether the goal was
realistic, and remembers the answer") - the reflection text read back empty.
The test clicked Save, then checked `expect(notices(page)).toHaveCount(0)`
before reloading. That assertion is true before the click as well as after
it - `.notice` only appears on an error - so it never actually waited on
anything; the reload could (and under load, did) run before the PATCH
landed. Reproduced deterministically by delaying the PATCH response 1.5s via
`page.route`: fails every time without a real barrier, passes every time
with one. Fixed by waiting on the PATCH response itself
(`page.waitForResponse`) before the reload.

**zz-library-missing.spec.js:153** ("a score whose file has gone is shown as
missing rather than vanishing") - an uploaded score "never appeared". This
is a genuine PRODUCT race, not a test bug. `scanner.start_scan()` is
fire-and-forget and declines outright when a scan is already running or the
library is mid-mutation - and until now, a decline was simply lost. If a
scan already in flight took its directory listing before a new upload's file
existed, and that upload's own `start_scan()` call was declined because the
other scan hadn't finished yet, nothing was left to ever look for the file
again - not eventually, not on the next scan, not ever, until something
unrelated happened to trigger one. Reproduced at the unit level: a real,
un-mocked scan made to take 0.3-1s before its listing returns, with a second
file landing mid-scan - the row for that file never appears. Fixed in
`server/fermata/scanner.py` with a `_rescan_pending` flag: a decline queues
exactly one follow-up scan, run automatically once the blocking scan or
mutation finishes. Pinned with two new tests
(`test_a_scan_declined_while_one_is_running_gets_a_follow_up_by_itself`,
`test_a_scan_declined_by_a_mutation_gets_a_follow_up_once_it_lifts`) that go
red when the fix is reverted. One existing test
(`test_a_scan_will_not_start_while_the_library_is_being_changed`) asserted
the old (buggy) behaviour - that `start_scan()` right after the exclusion
lifts returns `True` - which is now a race against the automatic follow-up;
updated to assert the actual contract (a scan runs, without anything asking
again) instead.

## The sweep

Every other browser spec (18 files) was swept for the same shape - a page
action followed by a read through the request context or a bare page read,
with nothing ordering the read after the write. All of it already follows
the correct #82/#100 idiom (a retrying assertion on something causally
downstream of the write, or an `expect.poll`/`.toPass()` block). No further
violations found. One unrelated pre-existing flake surfaced in a full-suite
run - `practice-shortcuts.spec.js:796` (gig-mode PDF page turn) - which is
already tracked separately in #168 (reopened, a genuine low-probability
product race in the PDF pane readiness path). Not #110's shape; not touched
here.

## The durable guard

#110 asks for more than another sweep - a point-in-time grep "does not
hold", because the wrong version is the natural thing to write. Added
`web/scripts/check-out-of-band-reads.mjs`: a crude, line-based scan of
`tests/browser/*.spec.js` (same spirit as the existing
`check-warning-patterns.mjs`) that flags a `request.*` read following a page
action with no intervening barrier, and wired it into `test:browser` so a
new instance fails fast, before a browser job has to flake under load to
surface it. Deliberately approximate - a false positive costs a
`// out-of-band-ok: <reason>` comment, which the one genuine false positive
this run turned up (a cleanup loop in metronome-everywhere.spec.js) now
carries. Mutation-tested against the original #82/#83 bug shape
(reinstating the ungated read makes it fail; restoring the barrier makes it
pass again).

Stated where the guard itself lives: it catches the request-context shape
specifically. It would NOT have caught practice.spec.js:343's flake, which
never touched `request.*` - the read channel was fine, the barrier
guarding a `page.reload()` was the defect. Telling that apart from the very
common and entirely correct "act, reload, assert" shape needs knowing which
barrier is causally tied to which write, which a text scan cannot know. A
green run of the guard is evidence against the request-context race, not
against every way a test can outrun its own write - said plainly rather
than implied, in the same spirit as this repo's own "say what the compiler
gate does not catch" precedent.

## Verification

- Both fixes mutation-tested: revert -> reproduces red; fix in place -> green.
- `server` suite: 1126 passed, 0 unexpected.
- `practice.spec.js` and `zz-library-missing.spec.js`, each run alone with
  `--repeat-each=20` under ~50% background CPU load: 300/300 and 60/60,
  0 unexpected, 0 flaky.
- The new guard mutation-tested against the reintroduced #82/#83 pattern.
- A full-browser-suite run is still completing locally at push time (CI will
  also run the full suite independently); will post the result here as a
  comment once it finishes. All targeted runs so far are green apart from
  the pre-existing, separately-tracked #168 flake noted above.
- No test counts changed; spec-floor guard untouched.

### DIFF
--- a/server/fermata/scanner.py
+++ b/server/fermata/scanner.py
@@ -98,17 +98,41 @@ TRASH_DIR_NAME = ".fermata-trash"
 #   inserts a row, or does not and the upload's own scan picks it up. Holding it
 #   would also break the ordinary case rather than protect it: every upload
 #   starts a scan, so uploading two files in a row would have the second refused
-#   for the first one's scan. The one interaction that matters is an upload
-#   landing on a move's destination, and that is caught where it has to be
-#   caught anyway (a person can drop a file into the library folder with no
-#   endpoint involved at all): _move_file_on_disk refuses a destination that
-#   exists, and its rename fails rather than overwriting if the file appears in
-#   the gap after that check.
+#   for the first one's scan - which is exactly why a decline queues
+#   `_rescan_pending` rather than dropping the request: without it, "the
+#   upload's own scan picks it up" was only true when that scan actually ran,
+#   and a scan already in flight when the second file landed had taken its
+#   listing before that file existed either (#110). The one interaction that
+#   matters is an upload landing on a move's destination, and that is caught
+#   where it has to be caught anyway (a person can drop a file into the
+#   library folder with no endpoint involved at all): _move_file_on_disk
+#   refuses a destination that exists, and its rename fails rather than
+#   overwriting if the file appears in the gap after that check.
 #
 #   CREATING A FOLDER IS NOT HELD either, and needs no argument beyond stating
 #   it: mkdir moves nothing and removes nothing, and a scan does not have an
 #   opinion about an empty directory.
 _mutating = False
+
+# Set when start_scan() is declined - because a scan is already running or
+# the library is mid-mutation - so the request it was declined for is not
+# silently lost (#110). The paragraph above claims "a scan either sees the
+# new file and inserts a row, or does not and the upload's own scan picks it
+# up" - that was only true when the upload's own scan actually runs. A scan
+# already in flight takes its directory listing before the new file exists,
+# so it does not see it either; if the upload's own start_scan() call is
+# THEN declined because that other scan is still going, nothing was left to
+# revisit the file at all. It stayed off the browser and out of every future
+# scan until something unrelated happened to ask again - seen as a browser
+# test's "the uploaded score never appeared" under CI load, where a scan
+# from a different test was still finishing when this one uploaded.
+#
+# The fix is not to make start_scan() block or queue every request - that
+# would turn "declined" into an ever-growing backlog under real load. One
+# flag is enough: the running pass (or mutation) is about to look at the
+# library fresh regardless, so anything that arrived while it was busy only
+# needs ONE more pass after it, not a pass of its own.
+_rescan_pending = False
 
 
 class LibraryBusy(RuntimeError):
@@ -149,6 +173,10 @@ def hold_library_still():
     finally:
         with _state_lock:
             _mutating = False
+        # If a scan was declined while this held the library (an upload
+        # landing mid-move, say), it does not get to fall through the gap
+        # the mutation just closed - see `_rescan_pending`.
+        _run_pending_rescan()
 
 
 def trash_dir(root: Path | None = None) -> Path:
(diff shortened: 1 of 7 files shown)


sirged:
Verification results, now complete:

**20-run flake check** (the reproduction condition both named tests shared),
each spec run alone with `--repeat-each=20` under ~50% background CPU load:

- `practice.spec.js`: 300/300 expected, 0 unexpected, 0 flaky
- `zz-library-missing.spec.js`: 60/60 expected, 0 unexpected, 0 flaky

**Full local browser suite** (`npm run test:browser`, includes the new
out-of-band-read guard as a pre-check): 508 passed, 0 failed.

**Full server suite** (`pytest`): 1126 passed, 0 unexpected (64 skipped for
want of a real sheet-music library fixture, as usual locally).

No occurrence of the pre-existing, separately-tracked #168 flake
(`practice-shortcuts.spec.js:796`) in any of these runs, though that one is
known to be load-marginal and outside this PR's scope regardless.

Issue #239, opened 2026-09-04
Cleanup and preparation for shipping Rev 1 to the server stack


sirged:
**Problem.** Rev 1 is about to move to the owner's server stack. Cycle 1's reviews left a short list of measured loose ends that are cheap now and expensive once deployed: the server CI job suffers the same 7-minute registry stall the browser job did and has no timeout; the browser-test runner still has one shell spawn; two needs-repro issues are unresolved (#226, #234); and several docs describe the pre-Cycle-1 product. Interpretation of "cleanup and prep for the next phase" by the dispatcher; correct it if the phase means something else.

**Appetite.** S to M, three agent-days.

**Solution sketch.**
- CI: cache the server job's `node_modules` with its OWN key (it installs `--omit=dev`; sharing the browser job's key would poison that tree), add a timeout floor to the server job, and add the cheap `--list` regression check for a two-word `--grep` to the Frontend build job (0.8 s, no browser). Drop the last `shell: true` spawn in `web/scripts/run-browser-tests.mjs`.
- Reproduction gate: reproduce #234 (`practice-shortcuts.spec.js` gig-mode band) with `--repeat-each 20` under 8x and 20x throttle on main and fix the test's measurement barrier or its band with a stated reason; reproduce #226 only as a read-only probe and either promote it to a bug with a failing spec or close it with the measurement.
- Docs: a second accuracy pass over `README.md`, `docs/deployment.md`, `docs/api.md`, `docs/practice-data.md` against current main (the MCP flag, the new score columns, the auto-transcription, the compose example), with the same claims-table discipline as #216.
- Repo: confirm `.gitignore` covers generated test output; confirm no library path or title has crept back into tracked text (re-run #221's broad sweep).

**Rabbit holes.** Rewriting CI beyond the two jobs named. Fixing #226 for real (a per-part ordinal map is a bet of its own).

**No-gos.** No feature work. No changes to what the suites test, only to how they are run and documented.

**Done when.** Server job shows a cache hit on its second run and has a timeout; the `--list` check fails when the runner is reverted to the shell spawn; #234 is closed by a fix with `--repeat-each` evidence or by a measured non-reproduction; #226 is a bug with a red spec or closed with the probe's numbers; the docs pass lands with a claims table and zero false claims; the title sweep returns only the conftest table.

sirged:
All three halves landed: #240 (CI), #241 + #248 (docs), #246 (reproduction gate). Done-when, item by item:

[...]

PR #240, opened 2026-09-04
Cache the server job's node_modules, floor its runtime, and guard the runner's argument passing

sirged:
Part of #239 (the CI half of it). Separate from the reproduction gate and the docs pass, which are their own PRs.

## Why

The server job installs `web/node_modules` with `--omit=dev` so an extracted transcription is parsed back by the real alphaTab importer instead of skipping. That one step is the whole of the job's runtime variance. Measured over recent `main` runs:

| Run | Job total | `Install alphaTab` | pytest |
| --- | --- | --- | --- |
| 33887700373 | 1.7m | 0.0m | 1.3m |
| 33841324577 | 6.4m | 5.1m | 0.9m |
| 33845939449 | 8.2m | 7.0m ("added 6 packages in 7m") | 0.8m |
| 33849328258 | 8.1m | 7.0m | 0.8m |
| 33854927005 | 8.7m | 7.0m | 1.2m |

pytest stays at 0.8-1.3 minutes throughout. Everything else is the registry stall the browser job already fixed for itself by caching its installed tree.

## What changed

**`.github/workflows/ci.yml`, server job.**

- Caches `web/node_modules` under `node-modules-omit-dev-${{ runner.os }}-node22-${{ hashFiles('web/package-lock.json') }}` and skips `npm ci --omit=dev` on a hit. The `omit-dev` segment is load-bearing: this job saves a tree with no playwright and no build tooling in it, and if it shared the browser job's key then whichever job ran first would decide what the other one restored - a browser job that restored this tree would skip its own `npm ci` and then have no playwright to run the suite with. Lockfile hash and Node major are there for the same reason as in that job.
- `timeout-minutes: 20`, as a floor under the cache rather than a substitute for it: 2.3x the worst run observed above, so a stall that outlives the cache fails in minutes instead of holding a runner for the six-hour default.

**`.github/workflows/ci.yml`, Frontend build job.** A sub-second regression check on the runner's own argument handling:

```
node scripts/run-browser-tests.mjs tests/browser/practice-shortcuts.spec.js --grep "gig mode" --list
```

must print exactly `Total: 6 tests in 1 file`. `--list` starts no browser and no server, so it needs only the `node_modules` and `web/dist` this job already has, and it costs under a second.

**`web/scripts/run-browser-tests.mjs`.** The last `shell: true` spawn - the out-of-band-reads guard - now uses `process.execPath` with no shell, like the Playwright spawn beside it. Its arguments are fixed and space-free so nothing was broken there; the point is that "this script never invokes a shell" is now true of the whole file rather than of one call in it, and the guard runs under the same node binary as the script instead of whichever one a PATH lookup finds.

## Evidence that the new check can fail

The check exists because the runner used to hand argv to a shell, and on Windows a shell joins argv into a single command line whose quoting does not cover internal spaces - so `--grep "gig mode"` arrived as `--grep gig` plus a stray `mode`, and Playwright ran a wider set than the caller asked for, green. Reverted the Playwright spawn to that exact form locally:

```js
const result = spawnSync("npx", ["playwright", "test", `--reporter=${reporters}`, ...filteredArgs], {
  cwd: webRoot, stdio: "inherit", shell: true, env: { ... },
});
```

and re-ran the check:

```
  ... practice-shortcuts.spec.js:919:3  > ... is honoured in gig mode too, where a pedal tap comes from
  ... practice-shortcuts.spec.js:1049:3 > ... a half-page gig turn pressed mid-re-render is honoured, not rolled back
  ... practice-shortcuts.spec.js:1111:3 > gig mode itself > from the default (side-by-side) layout, ...
  ... practice-shortcuts.spec.js:1190:3 > gig mode itself > a half-page turn survives the pane being re-rendered ...
  ... practice-shortcuts.spec.js:1254:3 > gig mode itself > the page indicator matches the pane ...
  ... practice-shortcuts.spec.js:1290:3 > gig mode itself > a resize that changes which page is on screen ...
  ... practice-shortcuts.spec.js:1325:3 > gig mode itself > from the staff layout, gig mode keeps the staff pane ...
Total: 7 tests in 1 file
```

Seven, not six - the one extra is the test whose title contains "gig" but not "gig mode", which is exactly the degradation. Restored the spawn; the same command lists six and the step passes. The counts are not arbitrary: 6 composed titles in that file match "gig mode" (one directly, five through their describe block), 7 match "gig".

After review, the step also pins the wider count: `--grep "gig" --list` must print `Total: 7 tests in 1 file`. Without that, renaming or removing the one extra test would make both listings six and leave the step green while it proved nothing.

`PLAYWRIGHT_ALLOW_PARTIAL=1` is set on that step because `--list` executes nothing, so every spec file reports zero tests run and the runner's spec-floor guard would otherwise, and correctly, read the whole suite as having gone missing. Confirmed: without it the step fails with 60-odd "only 0 ran" lines.

## Also verified locally

- `web/scripts/run-browser-tests.mjs` has no `shell:` option left anywhere (only the comments explaining why).
- A real run still works through the changed guard spawn: `node scripts/run-browser-tests.mjs tests/browser/version.spec.js` - 2 passed, exit 0, "out-of-band read check passed (30 browser spec files)".
- `ci.yml` parses; the server job's steps are in the order cache, Install, Install alphaTab, Fetch schema, Test, and the web job's are Install, Build, Check.

## Cache behaviour to look for on this PR

The first run populates the server job's cache (a miss, `Install alphaTab` runs); a second run on the same lockfile should hit it and skip that step. Reported in a comment below once both runs exist.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,14 +118,21 @@ jobs:
       # playwright.config.js refuses to load without it).
       #
       # What it guards: run-browser-tests.mjs used to hand its argv to a
-      # shell, and on Windows a shell joins argv into a single command line
-      # whose quoting does not cover internal spaces. A multi-word --grep
-      # came out the far side as separate bare words, so Playwright saw only
-      # the first one and ran a wider set than the caller asked for -
-      # silently, with a green run to show for it. Six tests in this file
-      # have "gig mode" in their title and seven have "gig", so a --grep that
-      # loses its second word lists seven and this step goes red. Proven
-      # against that exact revert before it was added.
+      # shell, and a shell spawn joins argv into one unquoted command line
+      # (on every platform, not only Windows), so a multi-word --grep came
+      # out the far side as separate bare words. Playwright saw only the
+      # first one and ran a wider set than the caller asked for - silently,
+      # with a green run to show for it. Six composed titles in this file
+      # match "gig mode" (one directly, five through their describe block)
+      # and seven match "gig", so a --grep that loses its second word lists
+      # seven and this step goes red. Proven against that exact revert
+      # before it was added.
+      #
+      # Both counts are pinned, not just the good one: if the titles ever
+      # converged so that "gig" and "gig mode" listed the same six, the
+      # first assertion would stay green while proving nothing. The second
+      # makes that convergence red instead of quiet. Adding or renaming a
+      # test whose title contains "gig" moves these numbers on purpose.
       #
       # PLAYWRIGHT_ALLOW_PARTIAL because --list executes nothing at all: every
       # spec file reports zero tests run, which the runner's spec-floor guard
@@ -139,6 +146,8 @@ jobs:
             tests/browser/practice-shortcuts.spec.js --grep "gig mode" --list)
           echo "$listed"
           echo "$listed" | grep -qx "Total: 6 tests in 1 file"
+          wider=$(node scripts/run-browser-tests.mjs             tests/browser/practice-shortcuts.spec.js --grep "gig" --list)
+          echo "$wider" | grep -qx "Total: 7 tests in 1 file"
 
   browser:
     name: Browser tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,15 @@ jobs:
   server:
     name: Server tests
     runs-on: ubuntu-latest
+    # Measured on main: pytest itself is 0.8-1.3 minutes, and the whole job
+    # has run anywhere from 1.7 minutes (33887700373) to 8.7 minutes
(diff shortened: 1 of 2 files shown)


sirged:
Both runs now exist on main.

| run | key `node-modules-omit-dev-Linux-node22-f85fc91…` | Install alphaTab | pytest |
| --- | --- | --- | --- |
| 33946463237 (first after merge) | not found, saved after the job | ran, 3 s | 1285 passed, 76 skipped |
| 33949194129 (later main run) | restored, 2 s | skipped | 1288 passed, 76 skipped |

The skip count is identical across miss and hit, so the restored tree carries the alphaTab package the gated tests import; the extra three passes are #245's new tests. The cache-hit branch of the server job is exercised and green.

Issue #165, opened 2026-08-29
A TAB-staff note emitted without its string crashes the renderer's paint


sirged:
Found during #151's playback work, present before it (confirmed with that change disabled): the committed `navigation.pdf` fixture's transcription makes the renderer throw during paint — `TabBarRenderer.collectSpaces` reads `spaces[tuning.length - str]` with `str` undefined, because the emitted notes carry no `<string>` element on a TAB-clef staff. Almost nothing of that fixture actually draws.

Rule 9 makes `<string>`/`<fret>` the tab profile's core pair, so notes emitted without them on a TAB staff are either an extractor gap on this fixture's shape or a fixture-generation gap — determine which first: does `tabextract` fail to assign strings for this score's notes, or does the emitter drop them?

## Verifying

- The fixture's emitted MusicXML carries `<string>` and `<fret>` on every TAB-staff note.
- The score page renders without console errors and the tab glyphs draw.
- A library sweep for any other transcription emitting stringless TAB-staff notes: report the count (expected 0 after the fix).

PR #304, opened 2026-09-15
Edit the bars, not only the notes in them

### LINKED ISSUE #300: The editor can change any note but not the bars around them: no way to set a key, a time signature, or add and remove measures
**Problem.** On main `33a0379`, every mutator `createDocument` returns is note-scoped and
addressed by note ordinal — `setFret`, `setString`, `setAccidental`, `cycleSpelling`,
`setDurationType`, `setDots`, `setTie`, `deleteNote`, `moveToVoice`, `restToNote`
(`web/src/lib/editor/document.js:1338-1361`). There is no mutator for anything above a note.

A person editing a transcription can change every note in a bar but cannot change the key that
bar is in, the time signature it counts against, or how many bars the piece has.
`<key><fifths>` is read (`document.js:151-173`) only to spell pitches; nothing writes it. No
`<time>` or measure-count writer exists at all.

**Appetite.** M (one week). The architecture is already the cheap kind for this: every edit goes
through `TabViewer.svelte` `applyEdit` (`:679-698`) as mutate-DOM → `doc.text()` →
`view.editor.reload(after)`, and undo is a full-text snapshot pushed on `undoStack`. A
measure-addressed mutator therefore costs about what a note-addressed one costs, and inherits
undo, redo, the dirty flag, the refusal path and re-render without touching any of them. The week
is for the refusal rules and their tests, not the plumbing.

**Solution sketch.** `document.js` gains measure-addressed mutators beside the note-addressed
ones, each returning a boolean exactly as the existing ones do (`applyEdit` already reads a false
return as a refusal and surfaces the message):

- `setKey(measureIndex, fifths)`
- `setTime(measureIndex, beats, beatType)`
- `insertMeasure(afterMeasureIndex)`
- `deleteMeasure(measureIndex)`

Each writes the measure's own `<attributes>` child, created in schema order if absent, reusing
`firstChildTag` / `tagText` and `renumberMeasure`. `TabViewer.svelte` wires each through the
existing `applyEdit(mutate, refusal)` and adds a bar-scoped control group to the edit panel,
enabled when the selection resolves to a measure — `stepMeasure` already navigates them.

**Rabbit holes.**

- *Re-barring on a time-signature change.* Changing 4/4 to 3/4 makes existing bars violate Rule 8
  (every voice's notes and rests sum to the measure), which `document.js` keeps by construction
  (`:1072-1077`). Re-flowing content across barlines is its own bet. **Decision:** `setTime`
  refuses where a measure's content does not already sum to the new signature, and the refusal
  names the bar and both sums; it applies where the content does fit. A person re-bars by editing
  durations first.
- *Re-spelling on a key change.* `fifths` drives `spellPitch` / `keyAlter` (`:561-573`).
  **Decision:** `setKey` writes `<fifths>` and nothing else. Notes keep the spelling they carry;
  `cycleSpelling` is already how a person changes one.
- *An inserted bar's content.* **Decision:** an inserted measure is one whole-measure rest per
  voice present in the preceding measure, built with the existing `makeRest` and the same
  per-voice fill the measure rebuild already performs.
- *Attributes inheritance.* MusicXML carries key and time forward until restated, so a change at
  measure N changes the meaning of every later measure that does not restate its own.
  **Decision:** write at the selected measure only; a later measure that restates its own key or
  time is left untouched. The bet does not add a "change from here on" mode.
- *Deleting the last measure, or the only one.* **Decision:** `deleteMeasure` refuses on a
  single-measure document, with a reason.

**No-gos.**

- Clef editing. On a tab staff it means nothing, and these documents are tab-first.
- Re-flowing note content across barlines when the time signature changes (see rabbit hole one).
- Structure editing on PDF-backed scores; the editor is MusicXML-backed.
- A from-scratch blank document — still blocked on `scores.path NOT NULL`.
- Auto-save. The explicit save and the `dirty` flag stay exactly as they are.
- Any change to how transcriptions are saved; no precondition or version-check work rides along.

**Done when.** In the editor on a MusicXML-backed score, a person can select a bar and:

1. change its key signature, and the render shows the new key with no note's printed pitch
   changed;
2. insert a bar after it and see an empty bar appear, then delete that bar again;
3. change the time signature where the bar's content fits it, and be refused with a reason naming
   the bar and both sums where it does not.

Undo returns each of those in one step. The saved document re-imports, and the bar counters in
the transcription disclosures agree with the new structure.

**Review.** Adversarial review is a hard requirement here, not optional: this changes a
document-wide invariant (Rule 8) and touches attributes that carry forward across measures.

sirged:
Closes #300

The editor could change any note in a bar and nothing about the bar: no key
signature, no time signature, no way to add or remove one. Every mutator
`editor/document.js` offered was addressed by note ordinal; nothing wrote
`<key>`, `<time>` or the measure count at all.

*(Body rewritten at the third review round. It describes the branch as it
stands, not as first proposed; the review history is in the comments.)*

## What this adds

`document.js` gains four measure-addressed mutators beside the note-addressed
ones — `setKey`, `setTime`, `insertMeasure`, `deleteMeasure` — each writing the
selected measure's own `<attributes>` and returning a boolean the way the
existing ones do. `TabViewer` wires each through the **same** `applyEdit` the
note edits use, so undo, redo, the dirty flag, the refusal message and the
re-render are the ones already there. `applyEdit` grows one branch, not a
second path: a structural edit rebuilds the model from the new text (every list
`createDocument` reads is computed once at parse, so an insert or a delete
leaves them describing the previous document) and re-resolves the selection
against it.

The rebuild is attempted **before** anything is committed. A structural mutator
writes into the live document, so a result the model cannot be rebuilt from has
already been written by the time that is discovered; on a throw the pre-edit
text is put back (known-parseable — it is what the model was built from), the
selection is re-resolved and the edit is refused, naming what was wrong.
Nothing reaches the undo stack, which is what keeps `restore()`'s "a snapshot
always parses" true; its comment now states that invariant instead of resting
on "our own serializer produced it", which was never the reason.

The controls sit on their own full-width line of the edit panel, present and
disabled rather than absent when nothing is selected — see "the panel's height".

## The decisions, each a refusal rather than a repair

- **A meter the bar's music does not already add up to is refused**, naming the
  bar and saying how many beats it holds against how many the meter asks for:
  *"Bar 1 holds 4 quarter-notes of music and 3/4 holds 3."* Nothing is
  re-barred. The check spans the bars the change actually governs — the
  selected one, and every following bar that states no meter of its own. That
  is wider than the issue's literal "a measure's content", deliberately: Rule 8
  is a property of the whole document, and a change at bar 1 is what would
  break it at bar 2. It only ever refuses more; it never reflows anything. A
  bar that restates its own `<time>` is left out of the check and out of the
  change.
- **A key change writes `<fifths>` and nothing else.** No note is re-spelled.
- **An inserted bar** is one whole-bar rest per voice the preceding bar is
  written in, built with the existing `makeRest`, so it satisfies Rule 8 by
  construction. It states no attributes of its own.
- **The last remaining bar cannot be deleted.**
- **A bar carrying a repeat or a volta ending cannot be deleted**, naming it. A
  `<barline>` is a child of `<measure>`, so a deletion takes it with the bar,
  and the surviving half of a repeat pair then means "repeat from the start of
  the piece" — silently changing the form of every bar before it. No barline
  migration is attempted. A plain final barline carries neither and stays
  deletable, with its own test so the refusal cannot quietly widen.
- **Deleting a bar carries its `<attributes>` onto the bar that follows**,
  except what that bar declares itself. Without it, deleting the first bar
  takes the document's divisions, clef and staff tuning with it. `<key>`,
  `<time>` and `<clef>` are all-or-nothing — each is one complete statement a
  restatement replaces — while `<staff-details>` is merged into, because it
  describes the instrument and can legitimately be restated in part. Inside it
  a `<staff-tuning>` is matched by the `line` it tunes, not by its tag: there
  is one per string, and a bar restating two of six (a scordatura, a dropped
  bottom string) has not restated all six. A migrated tuning lands among the
  restated ones in `line` order, so a merged `<staff-details>` still runs from
  the lowest string up the way the profile writes it.
- **A structural edit renumbers the bars into one consecutive run** and
  re-derives the Rule 17 ids of every bar whose number moved, so two bars can
  never share a number. The run starts at the number the document already
  started at — an excerpt numbered from 40 stays numbered from 40 — except that
  deleting a pickup numbered 0 starts the run at 1.
- **A multi-note range is collapsed** by a structural edit, and only by one: a
  range is a span of ordinals and an insert or a delete renumbers them. An undo
  restores the very document a range was made in, so it keeps its range.

## Evidence

22 tests in `tests/browser/score-editor-structure-300.spec.js` and one added to
`score-editor-range-251.spec.js`, each watched red against the change it
covers. The first thirteen were watched red against this branch's own source
before any of it existed (`git stash` of `web/src`: 13 failed).

One of those assertions needed its fixture changed to be falsifiable at all: the
partial-restatement test asserted the merged tunings run `1..6`, but restated a
contiguous `1,2`, and four tunings appended after `1,2` land in order whether
anything ordered them or not. It restates `2,4` now, so removing the ordering
branch reds it. The consequence of that branch is profile conformance and diff
readability only — repeats of one tag are not ordered by `xs:sequence`, the
model reads tunings into a map keyed by `line`, and the render is unchanged — so
that single assertion is the only thing holding it.

Mutation-tested, each applied to a committed tree and reverted:

| Mutation | Result |
|---|---|
| Rule 8 sum check removed from `timeChangeRefusal` | 2 red; the "does fit" case still green |
| meter span narrowed to the selected bar | 1 red — exactly the inherited-bar test |
| `setKey` also re-spells the bar's notes | 1 red — "no printed pitch changes" |
| `renumberMeasures()` dropped after an insert | 3 red |
| attributes migration dropped from `deleteMeasure` | 1 red |
| structural model rebuild dropped from `applyEdit` | 3 red |
| `<staff-details>` skipped instead of merged into | 1 red |
| `createDocument` allowed to throw out of `applyEdit` | 1 red |
| `<staff-tuning>` presence keyed by tag, not by `line` | 1 red — `stringCount` 6 → 2 |
| range collapse dropped from the structural branch | 1 red; the undo test stays green |
| range collapse moved back onto the shared rebuild path | 1 red — the undo test; the structural test stays green |
| repeat/ending check dropped | 1 red; "plain final barline" stays green |
| renumbering pinned to 1 | 1 red — the excerpt test |
| beat count re-derived from `<divisions>` | 1 red — reproduces *"holds 4 divisions … and 6/8 holds 6"* |
| refusal reported in divisions again | 2 red |
| meter control reset dropped | 1 red — control reads "3/4" on a 4/4 document |
| `line`-order branch dropped from `insertInOrder` | 1 red — tunings serialise `2,4,1,3,5,6`; nothing else in that test moves |

Suites, on the merged head (`origin/main` merged in):

- One ordered invocation of every `score-editor*` spec plus `score-multi-part`,
  `viewer-practice`, `zz-library-missing` and `zzzzzzz-library-scan-poll` (the
  downstream specs that assert emptiness): **121 passed, 0 failed, 0 skipped**.
  This spec stubs the HTTP transport and creates no rows, like every other
  editor spec.
- `tests/unit`: 331 passed, 0 failed, 0 skipped.
- Server suite: no server file is touched by this branch.

The saved document re-imports: a test saves a key change, a meter change and an
inserted bar, reloads the page, and reads the persisted document back through
the real client. The transcription disclosures state "not recorded" for every
bar counter after any hand edit, by design
(`test_an_edit_states_that_nothing_measured_it`), so no counter can contradict
the new structure.

## The panel's height

`.edit-panel` is a wrapping flex ROW. A bar row that merely followed the note
fields shared a line with them in one selection state and took a line of its
own in another, so the panel's height changed with the selection — measured at
+47px — which moves the staff under a note-head already aimed at, and reds
`score-editor-poly-189`'s click-cycling tests. The row is pinned to a
full-width line of its own in every state; the height now changes by 2px across
a selection, the same as before this change.

`score-render.js` gains a read-only `barCount`/`barInfo` beside `viewInfo`:
alphaTab's own read of the re-imported bars, so a structural assertion has
independent evidence that the edit reached the screen and not only the
document.

`docs/musicxml-tab-profile.md`'s Rule 4 now states what the editor may write
after the fact, what a deletion moves, how `<staff-details>` and its tunings
are merged, and that a `<barline>` is never moved.

Adversarial review is a hard requirement on this one — it changes a
document-wide invariant.


sirged:
All five findings fixed in `dfa33f4`, plus the refusal wording. Every fix has a test watched red against that fix and green with it.

**F1 — a bar edit that leaves an unopenable transcription.** Both halves fixed, and the general one is the primary fix.

`applyEdit`'s structural branch now attempts the rebuild **before** anything is committed. On a throw it re-parses the pre-edit text (known-parseable — it is what the model was built from), re-resolves the selection and refuses the edit, naming what was wrong. Nothing is pushed onto the undo stack and `dirty` is untouched, so `restore()`'s claim that a snapshot always parses stays true; its comment now states the invariant that keeps it true instead of asserting it as if it were free.

The migration now merges `<staff-details>` into the one the next bar declares, reading the target's tags **once** so a repeating child (six `<staff-tuning>`s) moves as a set rather than the first one making the rest look present. `<key>`, `<time>` and `<clef>` stay all-or-nothing on purpose: each is one complete musical statement and a bar restating one replaces it; `<staff-details>` describes the instrument and can legitimately be restated in part.

Testing the guard needed a document where the result is unopenable *after* the migration is correct — the tuning declared in the last bar, which has no later bar to migrate to. That staff draws no note heads (the unstrung-tab guard, #165, drops the bounds), so that one test selects through the model harness rather than by clicking, and says so; every assertion in it is on the document and the panel.

**F2 — the range.** `reselectAfterRebuild` collapses to the anchor. Measured before the fix, exactly as described: `data-editor-selected-count` read 2 over renumbered survivors.

**F3 — repeats.** `deleteMeasureRefusal` refuses a bar carrying a `<repeat>` or an `<ending>`, naming the bar and which. No barline migration attempted. A plain final barline carries neither and stays deletable — that has its own test, so the refusal cannot quietly widen into "any bar with a barline".

**F4 — the meter control.** Put back in step with the document after every attempt, refused or not, rather than only on refusal: the document is the source of truth here as everywhere else.

**F5 — excerpt numbering.** The run starts at the number the document already started at, captured before the change. Pickup handling stands: deleting a pickup numbered 0 starts the run at 1, because the bar that becomes first is a whole bar.

**Wording.** "Bar 1 holds 4 quarter-notes of music and 3/4 holds 3, so the time signature can't change here. Change the durations in that bar first." A bar that holds a fractional number of beats says 3.5 rather than rounding away the half beat that is the reason it does not fit. The spec pins the new wording and asserts the divisions figure is *not* in it.

**Mutations, all run on this commit** (each applied, measured, reverted):

| Mutation | Result |
|---|---|
| skip `<staff-details>` instead of merging into it | 1 red — "an attribute the next bar only partly restates" |
| let `createDocument` throw out of `applyEdit` | 1 red — "would leave an unopenable transcription is refused" |
| drop the range collapse | 1 red — selected-count 2 instead of 1 |
| drop the repeat/ending check | 1 red — the repeat test; "plain final barline" stays green |
| pin renumbering to 1 | 1 red — the excerpt test; "numbered run" stays green |
| report the sums in divisions again | 2 red — both refusal tests |
| drop the meter control reset | 1 red — control reads "3/4" on a 4/4 document |

The first commit's six mutations were re-run unchanged and still red.

**Suites on this commit.** The ordered invocation (every `score-editor*` spec + `score-multi-part` + `zz-library-missing` + `zzzzzzz-library-scan-poll`): **110 passed, 0 failed, 0 skipped** — 103 before, plus the 7 new tests. `tests/unit`: 331 passed, 0 failed, 0 skipped. Spec floor raised 13 → 20 in my own entry.

Two things I deliberately did not do, as asked: nothing added to the vocabulary guard's component list, and the two pre-existing items (undo-to-original leaving `dirty` true, the narrow-width staff shift identical on main) are untouched.

sirged:
D1, D2 and D3 fixed in `1995c8f`; `origin/main` merged in and the whole thing re-run on the merged head. The PR body is rewritten to match the branch as it stands.

**D1 — the tuning loss, one level down.** Presence in the `<attributes>` merge is no longer a set of tag names: `sameChildKey` keys a `<staff-tuning>` by the `line` it tunes, because there is one per string and two of six is not a restatement of six. A migrated tuning is also inserted among the restated ones in `line` order rather than after all of them, so the profile's "lowest string first" stays true through a merge. `<key>`, `<time>` and `<clef>` stay all-or-nothing — each is one complete statement a restatement replaces.

I considered whether the migration can be made safe at this size, since that was the line you drew, and my answer is yes rather than "refuse instead": the failure mode both times was the same one — matching by a coarser key than the thing being matched — and it is now keyed by what actually identifies the element. The recursion is one level deep and bounded by `STAFF_DETAILS_ORDER`, and the F1 rebuild guard sits underneath as the backstop for anything that still slips. If you would rather have `deleteMeasure` cut from the bet anyway, say so and I will pull it out; the other three mutators do not depend on it.

**D2 — the regression I introduced.** The collapse moves out of `reselectAfterRebuild` (which undo's `restore` also calls) into `applyEdit`'s structural branch, where the reason for it lives. `score-editor-range-251.spec.js` gains the test that was missing: a range survives a plain undo with the same anchor, extent, count, ordinals and marked heads — and still *works*, so the next range operation acts on all three notes rather than the anchor alone. Its floor entry is a new file, not an edit to yours.

**D3 — the divisions leak.** The beat is taken from the meter's own measured length (`expected / beats`), not re-derived from `<divisions>`, so it is exact, never missing, and may be a fraction of a division. With `<divisions>1</divisions>` and 6/8 the bar now reads *"holds 8 eighth-notes of music and 6/8 holds 6"*. The two-decimal trimming is untouched. The separate "this document's divisions cannot express 7/8" message is unchanged and still names divisions, because that one is about them — the new test pins both sentences on the same document.

**Nits, since I was in the files:** the insert refusal now says what it is refusing, and the two double blank lines are gone.

**Mutations watched this round:**

| Mutation | Result |
|---|---|
| `<staff-tuning>` presence keyed by tag, not by `line` | 1 red — `stringCount` 6 → 2, your measurement exactly |
| collapse moved back onto the shared rebuild path | 1 red — the new undo test (extent "" where "5" is wanted); the structural collapse test stays green |
| collapse dropped from the structural branch | 1 red — the structural test (count 2, not 1); the undo test stays green |
| beat re-derived from `<divisions>` | 1 red — reproduces *"Bar 1 holds 4 divisions of music and 6/8 holds 6"* |

The two collapse mutations are the pair that pins D2 from both sides: neither test alone would have caught the regression.

**Suites on the merged head:** the ordered invocation (every `score-editor*` spec + `score-multi-part` + `viewer-practice` + `zz-library-missing` + `zzzzzzz-library-scan-poll`) **121 passed, 0 failed, 0 skipped**; `tests/unit` 331 passed, 0 failed, 0 skipped. Floors: structure 20 → 22, plus a new 1-test entry for range-251.

